PR #2, opened 2020-06-29
Update to version v20.1.3

asubiotto:


Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
 version: 4.0.0
```
jlinder, line 4: This also needs to be bumped to `4.0.1`. 

It's the version of these cockroachdb charts and it needs to be unique for each change we make.

asubiotto: Done.

PR #3, opened 2020-06-29
Update the readme to reflect the new repository location

jlinder:
Before: The readme described how to work with the old repository
location

Why: We want users to install from the new repository.

Now:

- All references to the chart repository now point to the new repository
- Instructions for how to add the repository to Helm are included


jlinder:
Ah! I did not realize that. Done.

PR #4, opened 2020-07-01
Automate linting with TeamCity

jlinder:
Before: Lint wasn't being run automatically for every PR

Why: So we can ensure the charts are passing lint

Now:

- add scripts to run `ct lint` in a docker container.
- move the cockroachdb chart to the `cockroachdb` directory so ct picks
  up the chart correctly.
- update the chart to pass lint.


build/tester.sh
```diff
@@ -0,0 +1,42 @@
...
+# Absolute path to the toplevel helm-charts directory.
+helm_charts_toplevel=$(dirname "$(cd "$(dirname "${0}")"; pwd)")
+
+container_home=/charts/
+
+# Run our tester container with a set of volumes mounted that will
+# allow the container to store persistent data on the host computer.
+
+vols=
+vols="${vols} --volume=${helm_charts_toplevel}:${container_home}"
```
DuskEagle, line 31: Why not make this a single line?

`vols=--volume=${helm_charts_toplevel}:${container_home}"`

jlinder: That was vestigial from the script I used as a template for this one and I hadn't changed it. Updated to just one line.

jlinder:
TFTR!

PR #12, opened 2020-07-07
Automate the charts release process

jlinder:
Before: releasing was manual toil

Why: to make releasing repeatable, less error prone and easy

Now:

Adding files for building and publishing new charts. This supports the
"Make and Publish Charts Release" TeamCity build configuration.

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+helm_charts_toplevel="$(dirname "$(cd "$(dirname "${0}")"; pwd)")/"
+builder="${helm_charts_toplevel}/build/builder.sh"
+charts_hostname="${CHARTS_HOSTNAME:-charts.cockroachdb.com}"
+
+if [ -n "${DISTRIBUTION_ID-}" ] ; then
+  distribution_id="${DISTRIBUTION_ID}"
+elif [ "${charts_hostname-}" = "charts.cockroachdb.com" ] ; then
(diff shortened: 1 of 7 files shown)


build/release.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
+set -euox pipefail
```
DuskEagle, line 3: With `set -x`, I think this will cause the AWS_SECRET_ACCESS_KEY to be printed in the logs of the TC job.

I think something like this could work?

```
set -euxo pipefail

# some code

set +x
export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}" 
export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}" 
set -x
 
# more code
```

Also, minor nit: `pipefail` is an option name passed to the `-o` flag, so most scripts I've seen  put the `o` after the `e`, `u`, and `x`, e.g. `-euxo pipefail`.

jlinder: TeamCity is smart enough to realize when the characters of a secret are printed to the logs, it masks the value. [See this log for an example](https://teamcity.cockroachdb.com/buildConfiguration/Internal_Release_MakeAndPublishChartsRelease/2061098?buildTab=log&focusLine=261&linesState=196). That said, the x doesn't _need_ to be there and I've removed it.

jlinder:
@DuskEagle PTAL

jlinder:
TFTR!

PR #15, opened 2020-07-15
Adding cluster-name to init job

nyarly:
I found that without `--cluster-name` the init job would hang indefinitely if the cluster name was set in `values.yaml`.

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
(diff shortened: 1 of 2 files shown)


DuskEagle:
Thank you for fixing this @nyarly!

All updates to Helm charts require a version bump in [Chart.yaml](https://github.com/cockroachdb/helm-charts/blob/25e4598e13fa9402242d837e828a550a20e8a648/cockroachdb/Chart.yaml#L4). If you bump the version to 4.0.5 in that file, we can merge this.

nyarly:
@DuskEagle I updated against master, so this is actually 4.1.1

PR #19, opened 2020-08-11
Cert manager certificates

PaulFarver:
Originally https://github.com/helm/charts/pull/22902

#### What this PR does / why we need it:

- Add cert-manager certificate resources to the helm chart, to avoid users having to manually match subject alternative names in certificate resources
- Update documentation on setting up secure clusters

#### Special notes for your reviewer:

- The implementation of this is made with maximum backwards compatibility in mind. I've tested that the new templates render to the same result as the old ones, when using the same values. The right thing to do here, seems to be to rework the template a bit, but it'll break backwards compatibility, and require a major version bump.
- The implementation only covers the certificates, but it could be easily expanded to include the certificate authority and the issuer, with inspiration from https://github.com/helm/charts/pull/22838 

#### Checklist

- [x] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [x] Chart Version bumped
- [x] Variables are documented in the README.md

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 5.0.0
+version: 5.1.0
 appVersion: 20.2.0
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
 icon: https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png
--- a/cockroachdb/README.md
+++ b/cockroachdb/README.md
@@ -371,6 +371,10 @@ For details see the [`values.yaml`](values.yaml) file.
 | `tls.certs.clientRootSecret`                              | If certs are provided, secret name for client root cert         | `cockroachdb-root`                               |
 | `tls.certs.nodeSecret`                                    | If certs are provided, secret name for node cert                | `cockroachdb-node`                               |
 | `tls.certs.tlsSecret`                                     | Own certs are stored in TLS secret                              | `no`                                             |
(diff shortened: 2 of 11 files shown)


notetook:
When is it merged, and released? I'd like to use that 🥺

DuskEagle (commented):
I'm having a bit of trouble making this work. I've created a new self-signed Issuer like below:

```
apiVersion: cert-manager.io/v1alpha3
kind: Issuer
metadata:
  name: cockroachdb
spec:
  selfSigned: {}
```

I then use Helm to create a CockroachDB cluster which uses `cert-manager`. Helm creates the cluster just fine,  but the `cockroachdb-init` job is unable to connect to any of the `cockroachdb` pods in the cluster. It instead reports the error:

```
warning: node not ready to perform cluster initialization: initial connection heartbeat failed: rpc error: code = Unavailable desc = all SubConns are in TransientFailure, latest connection error: connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority
```

This suggests to me that the `root` client cert or secret being created doesn't quite work for initializing the cluster. Do you have any further guidance on how to make this work?

PaulFarver:
@DuskEagle thanks for the reply 😄 

This iteration does not provision the certificate authority on it's own, so it relies on the provided issuer to have a certificate authority. In the case of a selfsigning issuer, that isn't the case. That points to a flaw in my design, I think.

The solution would be issue the certificate authority from the provided issuer, and then use that to issue the certificates.

I can cook it up tomorrow.

keith-mcclellan:
@DuskEagle what do you think about the current state here?  This could help us with the removal of the beta CA api that allows the `legacy-unknown` signer so I think we should update and merge.

DuskEagle:
I've opened #75 which rebases this PR on top of master and bumps the chart version accordingly.

PR #20, opened 2020-08-29
Apply the same env of statefulset in job.init

ribeiromiranda:
I needed to apply this COCKROACH_SKIP_KEY_PERMISSION_CHECK environment variable to both yaml.

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -118,6 +118,10 @@ spec:
               [[ "$initOUT" == *"cluster has already been initialized"* ]] && exit 0;
               sleep 5;
               done
+          env:
+          {{- with .Values.statefulset.env }}
+            {{- toYaml . | nindent 12 }}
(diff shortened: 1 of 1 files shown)


DuskEagle:
I think needing to use the environment variable `COCKROACH_SKIP_KEY_PERMISSION_CHECK` suggests your secrets aren't being mounted with read-only permissions. I think fixing that would be better than using this environment variable to bypass the issue. Let me know if I'm missing something.

ribeiromiranda:
The permission used was -r--r-----, but an error occurred while checking the permission.

DuskEagle:
>The permission used was -r--r-----, but an error occurred while checking the permission.

That is actually expected behaviour: `group` and `other` are not expected to have read, write, or execute permissions on your keys. See https://github.com/cockroachdb/cockroach/blob/b0b53c777f08721519f8f92d751652a22404a667/pkg/security/certificate_loader.go#L98

PR #21, opened 2020-08-31
expose topologyKey of pod anti-affinity rules as a variable

EladDolev:
very useful for small clusters, where number of replicas <= number of AZs

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 4.1.11
+version: 4.1.12
 appVersion: 20.1.8
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
 icon: https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png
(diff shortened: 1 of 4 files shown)


DuskEagle:
For my benefit, could you say more about how this helps when the number of replicas is <= number of AZs? This anti-affinity rule is meant to prevent two CockroachDB pods from being scheduled onto the same node. You may be planning to substitute `topology.kubernetes.io/zone` as the anti-affinity rule, but I wouldn't recommend that - it could prevent you from scaling up your cluster later. Instead, I would recommend relying on the Kubernetes scheduler to distribute each CRDB pod to a different AZ, unless there is a reason that doesn't work for you.

Separately, can you please add `statefulset.podAntiAffinity.topologyKey` to the [configuration section of the README](https://github.com/cockroachdb/helm-charts/tree/master/cockroachdb#configuration)?

EladDolev:
well when number of replicas is <= number of AZs I may substitute `topology.kubernetes.io/zone` and set `statefulset.podAntiAffinity.type` to `hard` thus ensuring each replica is running on a different zone (and a different host)
when number of replicas is > number of AZs, `statefulset.podAntiAffinity.type` must be `soft`, but it still gives a very good guarantee of spreading across AZs which IMHO is more important than having pods running on different nodes in the same AZ
from my knowledge and experience K8s will not necessarily distribute across AZs, especially in a large cluster.
I do agree this strategy is less effective for larger deployments

will make the adaptations, thank you !

EladDolev:
version bumped
appreciate it !

DuskEagle:
Getting a merge conflict on Chart.yaml trying to merge - can you pull the latest commits from master into your `topologykey` branch and push?

EladDolev:
rebased to latest master
thanks !

EladDolev:
Hi @DuskEagle !
Think we can merge this ?

DuskEagle:
Hi @EladDolev, sorry for not noticing your message earlier. It appears there's another merge conflict with cockroachdb/Chart.yaml. If you want to push another update I can merge this ASAP. Otherwise, I can open a fork of this PR with an updated Chart.yaml and merge that.

EladDolev:
updated
thanks !

PR #25, opened 2020-09-21
Fix #24

### LINKED ISSUE #24: Can't add items for NetworkPolicy
Reproduction:

```yaml
# config.yaml
networkPolicy:
  enabled: true
  ingress:
    grpc: 
    - podSelector:
        matchLabels:
          foo: bar
```

```
helm lint --debug -n cockroachdb  ./cockroachdb -f ./config.yaml 
==> Linting ./cockroachdb
[ERROR] templates/: template: cockroachdb/templates/_helpers.tpl:14:14: executing "cockroachdb.fullname" at <.Values.fullnameOverride>: can't evaluate field Values in type []interface {}

Error: 1 chart(s) linted, 1 chart(s) failed
helm.go:75: [debug] 1 chart(s) linted, 1 chart(s) failed
main.newLintCmd.func1
        /private/tmp/helm-20200213-73045-zsskjg/src/helm.sh/helm/cmd/helm/lint.go:113
github.com/spf13/cobra.(*Command).execute
        /private/tmp/helm-20200213-73045-zsskjg/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:826
github.com/spf13/cobra.(*Command).ExecuteC
        /private/tmp/helm-20200213-73045-zsskjg/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:914
github.com/spf13/cobra.(*Command).Execute
        /private/tmp/helm-20200213-73045-zsskjg/pkg/mod/github.com/spf13/cobra@v0.0.5/command.go:864
main.main
        /private/tmp/helm-20200213-73045-zsskjg/src/helm.sh/helm/cmd/helm/helm.go:74
runtime.main
        /usr/local/Cellar/go/1.13.8/libexec/src/runtime/proc.go:203
runtime.goexit
        /usr/local/Cellar/go/1.13.8/libexec/src/runtime/asm_amd64.s:135
```

travigd:
The use of `{{- with ... }}` seems useless here, and it breaks everything else. I think(?) it doesn't cause an issue in the default configuration because `.Values.networkPolicy.ingress.grpc` is an empty array (and I guess that means it doesn't actually switch into that scope?).

Would be nice™ to add a test too (I made this PR from the GitHub web app).

### DIFF
--- a/cockroachdb/templates/networkpolicy.yaml
+++ b/cockroachdb/templates/networkpolicy.yaml
@@ -23,10 +23,9 @@ spec:
   ingress:
     - ports:
         - port: grpc
-    {{- with .Values.networkPolicy.ingress.grpc }}
       from:
         # Allow connections via custom rules.
-        {{- toYaml . | nindent 8 }}
+        {{- toYaml .Values.networkPolicy.ingress.grpc | nindent 8 }}
         # Allow client connection via pre-considered label.
         - podSelector:
             matchLabels:
@@ -49,7 +48,6 @@ spec:
               {{- toYaml . | nindent 14 }}
             {{- end }}
       {{- end }}
-    {{- end }}
     # Allow connections to admin UI and for Prometheus.
     - ports:
         - port: http


EladDolev:
doesn't seem useless at all...
think the right fix here would be placing a $ sign before .Release in places they are inside a `{{- with...}}` scope
`.Release.Name` -> `$.Release.Name`

travigd:
Maybe I'm missing something, but it looks to me like the `with` is only being used to make `toYaml .` work - everything else in the block doesn't seem to depend on accessing anything inside `.Values.networkPolicy.ingress.grpc`.

But yeah, just using `$` to access the root object would also work here (it would just require changing more lines of code).

DuskEagle:
Good catch @travigd. I agree that the `{{- with ... }}` doesn't seem necessary, and that this is an easier change than replacing every templated call within the `with` scope to start with`$`.

If you increment the `version` field of chart.yaml as described in [CONTRIBUTING.md](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/CONTRIBUTING.md), I can merge this change.

PR #28, opened 2020-09-28
Add support for Pod Topology Spread Constraints

EladDolev:
We'd love having this for increased resiliency when losing an AZ, for both small or large CRDB clusters

### DIFF
--- a/cockroachdb/README.md
+++ b/cockroachdb/README.md
@@ -240,94 +240,99 @@ Note that there are some backward-incompatible changes to SQL features between v
 The following table lists the configurable parameters of the CockroachDB chart and their default values.
 For details see the [`values.yaml`](values.yaml) file.
 
-| Parameter                                | Description                                                     | Default                                          |
-| ---------                                | -----------                                                     | -------                                          |
-| `clusterDomain`                          | Cluster's default DNS domain                                    | `cluster.local`                                  |
-| `conf.attrs`                             | CockroachDB node attributes                                     | `[]`                                             |
-| `conf.cache`                             | Size of CockroachDB's in-memory cache                           | `25%`                                            |
-| `conf.cluster-name`                      | Name of CockroachDB cluster                                     | `""`                                             |
-| `conf.disable-cluster-name-verification` | Disable CockroachDB cluster name verification                   | `no`                                             |
-| `conf.join`                              | List of already-existing CockroachDB instances                  | `[]`                                             |
-| `conf.max-disk-temp-storage`             | Max storage capacity for temp data                              | `nil`                                              |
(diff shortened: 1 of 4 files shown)


cockroachdb/README.md
```diff
@@ -240,93 +240,97 @@ Note that there are some backward-incompatible changes to SQL features between v
...
+| `statefulset.priorityClassName`                           | [PriorityClassName][4] for StatefulSet Pods                     | `""`                                             |
+| `statefulset.tolerations`                                 | Node taints to tolerate by StatefulSet Pods                     | `[]`                                             |
+| `statefulset.topologySpreadConstraints`                   | [Topology Spread Constraints rules][5] of StatefulSet Pods      | auto                                             |
+| `statefulset.topologySpreadConstraints.maxSkew`           | Degree to which Pods may be unevenly distributed                | `1`                                              |
+| `statefulset.topologySpreadConstraints.topologyKey`       | The key of node labels                                          | `topology.kubernetes.io/zone`                    |
+| `statefulset.topologySpreadConstraints.whenUnsatisfiable` | `ScheduleAnyway`/`DoNotSchedule` for unsatisfiable constraints  | `ScheduleAnyway`                                 |
```
EladDolev, line 283: only these were added
the rest of the lines were just aligned

cockroachdb/templates/statefulset.yaml
```diff
@@ -128,6 +128,21 @@ spec:
         {{- end }}
       {{- end }}
     {{- end }}
+    {{- if semverCompare ">=1.16-0" .Capabilities.KubeVersion.Version }}
```
DuskEagle, line 131: I don't think this check is working as expected, based on the [test failure](https://teamcity.cockroachdb.com/viewLog.html?buildId=2326789&buildTypeId=Cockroach_HelmCharts_ChartsGitHubCi&tab=buildLog&_focus=465). `.Capabilities.KubeVersion.Version` [appears](https://github.com/helm/helm/blob/0ca95341c19fcaf9cecdf50caef1693568835a84/pkg/chartutil/capabilities.go#L48) to be a [k8s.io/apimachinery/pkg/version#Info](https://godoc.org/k8s.io/apimachinery/pkg/version#Info), which does not have a `Version` field.

DuskEagle (changes requested):

cockroachdb/templates/statefulset.yaml
EladDolev: changed to `.Capabilities.KubeVersion.GitVersion` which is deprecated on Helm v3 but still works :-)

DuskEagle (approved):
LGTM. I've updated the Chart.yaml version on this PR as well.

PR #30, opened 2020-10-12
don't specify max-disk-temp-storage and max-offset by default

DuskEagle:
When starting CockroachDB with `cockroach start(-single-node)`, we should not
be specifying the `max-disk-temp-storage` or `max-offset` flags unless the
user is intentionally overriding these values.

This actuallly resulted in some undesirable behaviour - our Helm chart was
setting the `max-disk-temp-storage` to zero bytes by default, which is not
the intended default value in CockroachDB. Previously, CockroachDB errroneously
reported zero bytes as the default value, but
[that has been fixed upstream](github.com/cockroachdb/cockroach/pull/54853).

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 4.1.9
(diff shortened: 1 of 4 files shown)


jlinder (approved):
LGTM

PR #35, opened 2020-10-27
Fixes cockroachdb/helm-charts#34

### LINKED ISSUE #34: Chart doesn't set requests/limits for "release-name"-init Job's own init container
Just as the title says. The root of the problem appears to be that the `job.init.yaml` template pulls in `init.resources` config from values pulls for the _containers_ config, but not for the _initContainers_.

`containers` stanza:

```
      containers:
        - name: cluster-init
          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
[...]
        {{- with .Values.init.resources }}
          resources: {{- toYaml . | nindent 12 }}
        {{- end }}
```

vs. `initContainers` stanza with no mention of `Values.init.resources`:

```
      initContainers:
        # The init-certs container sends a CSR (certificate signing request) to
        # the Kubernetes cluster.
        # You can see pending requests using:
        #   kubectl get csr
        # CSRs can be approved using:
        #   kubectl certificate approve <csr-name>
        #
        # In addition to the Node certificate and key, the init-certs entrypoint
        # will symlink the cluster CA to the certs directory.
        - name: init-certs
          image: "{{ .Values.tls.init.image.repository }}:{{ .Values.tls.init.image.tag }}"
          imagePullPolicy: {{ .Values.tls.init.image.pullPolicy | quote }}
          command:
            - /bin/ash
            - -ecx
            - >-
              /request-cert
              -namespace=${POD_NAMESPACE}
              -certs-dir=/cockroach-certs/
              -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt
              -type=client
              -user=root
          env:
            - name: POD_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
          volumeMounts:
            - name: client-certs
              mountPath: /cockroach-certs/
```

The stated goal of making these configurable values is to allow folks who are using resource quotas to avoid errors with the init job, but Resource Quotas apply to _all_ containers in a pod, including init containers. In order for this to work, we have to apply the logic to the initContainers as well.

rodalli:
Re-using the same chart logic that was used for the other init containers on the cluster-init job's init container in order to fix #34. I tested this on a 1.18 cluster with  resource quotas in place and it resolved the issue.

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -72,6 +72,9 @@ spec:
           volumeMounts:
             - name: client-certs
               mountPath: /cockroach-certs/
+        {{- with .Values.init.resources }}
+          resources: {{- toYaml . | nindent 12 }}
+        {{- end }}
     {{- end }}
     {{- with .Values.init.affinity }}
       affinity: {{- toYaml . | nindent 8 }}


rodalli:
@DuskEagle any ETA on when this can get merged? It's a very simple change and would help anyone using Resource Quotas in k8s a lot (being able to use the official CockroachDB chart vs. having to point to an altered one)

rodalli:
Found the same problem in the init container (init-certs) for the StatefulSet, so fixed that as well. It seemed like a good idea to allow users to control this with the values file just as with the init job's `init.resources` setting, so I created a similar block for `statefulset.init.resources` in `values.yaml` and commented that appropriately.

PR #36, opened 2020-11-06
chore(prometheus): Avoid tls problems related to pod IP.

semoac:
When using secure installations, and your own certificates, Prometheus will fail the scrapping because the pod IP was not in the original CSR.

The error on Prometheus looks like this:
```
Get https://10.6.26.160:8080/_status/vars: x509: certificate is valid for 127.0.0.1, not 10.6.26.160
```

This PR disabled the TLS verification, similar to what you need to do when setting other third parties integrations like Datadog's

Signed-off-by: Sergio Morales <sergio@cornershopapp.com>


cholick:
This was closed without merging? I'd run into this problem and was subscribed, hoping it would get to main. I don't see an alternative fix for https://github.com/cockroachdb/helm-charts/issues/37

Issue #45, opened 2020-12-08
Init job as a helm hook


sergeyshaykhullin:
For now init job is part of chart, deploys once and constantly available in a cluster

Main problem is to upgrade cluster, job is immutable and failing on crdb version upgrade. Workaround is to delete init job and helm upgrade again

I think that init job should be a helm hook like this (**IF INIT JOB IS IDEMPOTENT** and i think it is)

```yaml
annotations:
  helm.sh/hook: post-install,post-upgrade
  # This will delete job after succeeded
  helm.sh/hook-delete-policy: hook-succeeded,before-hook-creation
```

Minio implementation of buckets provisioning reference: https://github.com/minio/charts/blob/master/minio/templates/post-install-create-bucket-job.yaml

JorritSalverda:
When you run the job a second time it succeeds, even though the log shows:

```
ERROR: cluster has already been initialized Failed running "init"
```

That's good enough for the purpose of using the _helm hook_ and running the init job again for install and every upgrade. A `helm.sh/hook: post-install` without `post-upgrade` might also work, but this works for sure.

JorritSalverda:
Excellent. I verified that I can now upgrade the helm chart twice in a row. First time you probably still need to delete the init job manually with `kubectl delete job cockroachdb-init -n <namespace>`.

sergeyshaykhullin:
Hmm...

Issue #39, opened 2020-11-12
Upgrading helm chart causes error on cockroachdb-init job


Chili-Man:
I tried upgrading the Helm chart from 4.1.13 to 5.0.0 for updating Cockroach db to 20.2.0, but fails because it tries to update the Job resource, which is immutable. 

There's not really a good way to handle this in helm right now unfortunately due to the immutability aspect of it. The current workaround I have is to delete the job before the upgrade, but perhaps it would be better to make Job resource optional through a value like `init.enabled=false` so that it can be set on upgrades (I don't believe the init job needs to run on subsequent upgrades, but correct me if I'm wrong). 

The full error is below:

```console
rollback.go:83: [debug] updating status for rolled back release for cockroachdb
[...]

Issue #27, opened 2020-09-28
Init job breaks if you try to upgrade cockroach version


jimethn:
Jobs are immutable, but the init job uses the same values.yaml field (.Values.image.tag) as the actual DB pods. So when you update the version, helm (or rather, kube) errors out saying you're trying to change an immutable field.

jonasbadstuebner:
Should be fixed: https://github.com/cockroachdb/helm-charts/pull/68

Issue #7, opened 2020-07-03
Add database and users provisioning support


sergeyshaykhullin:
Add database and users provisioning like:
- users: https://github.com/zalando/postgres-operator/blob/88735a798a8c9ef98897ac94efa09a14527de83b/manifests/complete-postgres-manifest.yaml#L13
- databases: https://github.com/zalando/postgres-operator/blob/88735a798a8c9ef98897ac94efa09a14527de83b/manifests/complete-postgres-manifest.yaml#L22

e.g.
```yaml
users:
- username: username
  password:

databases:
- name:
  owners: [username]
```

sergeyshaykhullin:
@DuskEagle I want to start prototyping "provisioning" in a days

I my mind it will work as a post install helm hook, running job like init cluster, but running create if not exists statements. 

I don't think that values should be as complex as sql create constructions, so supports most used keywords for now

I can start with "raw sql" execution after cluster formation prototype

Any thoughts on this?

Chili-Man:
This feature would be great; we're currently thinking about doing it as a Kubernetes Job that can be ran as a separate release and depending on a flag e.g. `--set job.bootstrap` . I think changing the password/permissions/deletion should be out of scope and the concern should only be on bootstrapping dbs /users - there's no need to manage the entire lifecycle of those resources from this helm chart

keith-mcclellan:
I think we're more likely to do this in the operator rather than the helm chart.  We have an alpha operator available here: https://github.com/cockroachdb/cockroach-operator

sergeyshaykhullin:
```yaml
users:
- username: username
  password:

databases:
- name:
  owners: [username]
  # https://www.cockroachlabs.com/docs/stable/manage-a-backup-schedule.html
  backup:
    into: s3://test/backups/test_schedule_1?AWS_ACCESS_KEY_ID=123&AWS_SECRET_ACCESS_KEY=123
    with: [revision_history]
    schedule: @daily
    options:
      first_run: now
```

sergeyshaykhullin:
@keith-mcclellan No, i think that helm chart have to support users and databases with backups as a helm hook job described here https://github.com/cockroachdb/helm-charts/issues/45 (it can be separate), because CRDB supports `IF NOT EXISTS` for roles and databases and also native backup schedules <3

This features simplify helm structure to one helm hook job (no worries about cron jobs for backups, failed backups handing, jobs for provisioning, etc)

[...]

PR #46, opened 2020-12-10
Users, database, backups, organization and license provisioning support

sergeyshaykhullin:
Issue: https://github.com/cockroachdb/helm-charts/issues/7

## This pr also fixes (Init job is a helm hook now)
- https://github.com/cockroachdb/helm-charts/issues/45
- https://github.com/cockroachdb/helm-charts/issues/39
- https://github.com/cockroachdb/helm-charts/issues/27

## Questions:
- Does create schedule is idempotent?
- Existing database with different options?
- Existing user with different options?

## Tradeoffs
- I don't want to delete init job after its completion (it will can contain useful logs with errors), it will stay to next release and substitutes with new one, so if you uninstall helm chart you want to delete job manually, because it is hook, not a part of helm chart

## TODO
- [x] Check hook works, bash part
- [x] Check sql script works
- [x] Remove init parts? Remove while loop? Status codes?
- [x] Fix values for users and databases
- [x] Use commas around all database name usages?
- [x] Passwords and backup stores are not encrypted, i think i should create secrets and then map them to job
  - [x] Add secrets with user_PASSWORD?: "password"
  - [x] Pass to container with env vars secret ref
  - [x] Use ${{ $user.name }}_PASSWORD? env var
  - [x] Add same thing to license
- [x] Full backup schedule crontab?
- [x] Test multiple `create schedule` works (**IS NOT WORKING** https://github.com/cockroachdb/cockroach/issues/57892) **FIXED** in (https://github.com/cockroachdb/cockroach/pull/69152)
- [x] Test secure cluster with passwords

## Values
```yaml
init:
  # ...
  # TODO: dbms? cluster? provisioning? initialization? configuration?
  database:
    enabled: true
    organization: org
    license: license
    users:
    - name: test
      password: test
      options: [LOGIN]
    - name: test2
      options: [LOGIN, CREATEROLE]
    databases:
    - name: test
      options: [encoding='utf-8', connection limit=-1]
      owners: [admin]
      backup:
        into: s3:whatever
        options: [revision_history, encryption_passphrase=123]
        recurring: '@always'
        fullBackup: '@daily'
        schedule:
          options: [first_run = 'now']
    - name: test2
```

## Templating
```sh
helm template -f values.yaml test-release --debug --namespace test-namespace . > templated.yaml
```

```yaml
# Source: cockroachdb/templates/job.init.yaml
kind: Job
apiVersion: batch/v1
metadata:
  name: test-release-cockroachdb-init
  namespace: "test-namespace"
  labels:
    helm.sh/chart: cockroachdb-5.0.2
    app.kubernetes.io/name: cockroachdb
    app.kubernetes.io/instance: "test-release"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/component: init
  annotations:
    helm.sh/hook-delete-policy: before-hook-creation
spec:
  template:
    metadata:
      labels:
        app.kubernetes.io/name: cockroachdb
        app.kubernetes.io/instance: "test-release"
        app.kubernetes.io/component: init
    spec:
      restartPolicy: OnFailure
      terminationGracePeriodSeconds: 0
      containers:
        - name: cluster-init
          image: "cockroachdb/cockroach:v20.2.2"
          imagePullPolicy: "IfNotPresent"
          # TODO: COMMENT
          # Run the command in an `while true` loop because this Job is bound
          # to come up before the CockroachDB Pods (due to the time needed to
          # get PersistentVolumes attached to Nodes), and sleeping 5 seconds
          # between attempts is much better than letting the Pod fail when
          # the init command does and waiting out Kubernetes' non-configurable
          # exponential back-off for Pod restarts.
          # Command completes either when cluster initialization succeeds,
          # or when cluster has been initialized already.
          command:
          - /bin/bash
          - -c
          - >-
              initCluster() {
                while true; do
                  local output=$(
                    set -x;

                    /cockroach/cockroach init \
                      --insecure \
                      --host=test-release-cockroachdb-0.test-release-cockroachdb:26257
                  2>&1);

                  local returnCode="$?";
                  echo $output;

                  if [[ "$returnCode" == "0" || "$output" == *"cluster has already been initialized"* ]]
                    then break;
                  fi

                  sleep 5;
                done
              }

              initCluster;
              initDatabases() {
                while true; do
                  local output=$(
                    set -x;

                    /cockroach/cockroach sql \
                      --insecure \
                      --host=test-release-cockroachdb-0.test-release-cockroachdb:26257 \
                      --execute="
                          SET CLUSTER SETTING cluster.organization = 'org';
                          SET CLUSTER SETTING enterprise.license = 'license';
                          CREATE USER IF NOT EXISTS test WITH
                            PASSWORD '$test_PASSWORD'
                            LOGIN
                          ;
                          CREATE USER IF NOT EXISTS test2 WITH
                            PASSWORD null
                            LOGIN CREATEROLE
                          ;
                          CREATE DATABASE IF NOT EXISTS test
                              encoding='utf-8' connection limit=-1
                          ;
                          GRANT ALL ON DATABASE test TO admin;
                          CREATE SCHEDULE test_scheduled_backup
                            FOR BACKUP DATABASE test INTO 's3:whatever'
                            WITH revision_history,encryption_passphrase=123
                            RECURRING '@always'
                            FULL BACKUP '@daily'
                            WITH SCHEDULE OPTIONS first_run = 'now'
                          ;
                          CREATE DATABASE IF NOT EXISTS test2
                          ;

                        SELECT 'Provisioning completed successfully';
                      "
                    2>&1);

                  local returnCode="$?";
                  echo $output;

                  if [[ "$returnCode" == "0" || "$output" == *"Provisioning completed successfully"* ]]
                    then break;
                  fi

                  sleep 5;
                done
              }

              initDatabases;
          env:
          - name: test_PASSWORD
            valueFrom:
              secretKeyRef:
                name: test-release-cockroachdb-init
                key: test-password
```
</details>


sergeyshaykhullin:
I think that i should make init job as a helm hook with before-created delete policy and merge with job into init

So init will run each release (to deploy new crdb version now you have do same thing, but manually deleting job), init will try to initialize cluster first and then execute sql

Chili-Man:
@sergeyshaykhullin I think overall I like this approach; however, I think that the user passwords should come from a Secret resource instead of specifying directly as helm chart overrides (I see you've already mentioned this in your todo). To also support obtaining secrets from vault, it would be awesome to also allow the option of specifying the path to the secret file containing the secrets if a Kubernetes secret resource is not defined. 

For the values; I would consider putting it all under its own key , e.g. `initJob` so that it's a little clearer its part of the job. Adding a `initJob.enabled` flag should also be made to explicitly enable/disable the job, instead of implicitly like it's proposed now. 

Besides the user and database provisioning, it would be great if you can also add the   

```sql
SET CLUSTER SETTING cluster.organization = 'org';
SET CLUSTER SETTING enterprise.license = 'some secret';
```

for enterprise users as part of this initialization job (again the license key coming from a Secret resource / and or specified path (for vault)).

sergeyshaykhullin:
Yay, it works!

sergeyshaykhullin:
Secure cluster with passwords logs
```
2020-12-14T13:52:29.813171472Z ++ /cockroach/cockroach sql --certs-dir=/cockroach-certs/ --host=test-cockroachdb-0.test-cockroachdb:26257 '--execute=
2020-12-14T13:52:29.813215372Z             CREATE USER IF NOT EXISTS test2 WITH
2020-12-14T13:52:29.813224598Z               PASSWORD '\''test2'\''
2020-12-14T13:52:29.813250321Z               LOGIN CREATEROLE
2020-12-14T13:52:29.813258802Z             ;
2020-12-14T13:52:29.813266529Z             CREATE DATABASE IF NOT EXISTS test
2020-12-14T13:52:29.813274200Z                 encoding='\''utf-8'\'' connection limit=-1
2020-12-14T13:52:29.813288293Z             ;
2020-12-14T13:52:29.813295907Z             GRANT ALL ON DATABASE test TO test2;
2020-12-14T13:52:29.813303703Z             CREATE DATABASE IF NOT EXISTS test2
2020-12-14T13:52:29.813311381Z             ;
2020-12-14T13:52:29.813340936Z             GRANT ALL ON DATABASE test2 TO test2;
2020-12-14T13:52:29.813348859Z             GRANT ALL ON DATABASE test2 TO admin;
2020-12-14T13:52:29.813356581Z 
2020-12-14T13:52:29.813363896Z           SELECT '\''Provisioning completed successfully'\'';
2020-12-14T13:52:29.813372491Z         '
2020-12-14T13:52:31.104087173Z ?column? Provisioning completed successfully
```

sergeyshaykhullin:
I have no license, so i can't test its provisioning. Also full backup crontabs, if you will have time, can you help me with them?

sergeyshaykhullin:
I found a problem, schedules are not idemponent, so i can't run them multiple times
```
2020-12-14T14:37:57.991820027Z ++ /cockroach/cockroach sql --insecure --host=test-cockroachdb-0.test-cockroachdb:26257 '--execute=
2020-12-14T14:37:57.991863018Z             CREATE USER IF NOT EXISTS test WITH
2020-12-14T14:37:57.991872935Z               PASSWORD null
2020-12-14T14:37:57.991881374Z               LOGIN
2020-12-14T14:37:57.991889490Z             ;
2020-12-14T14:37:57.991897388Z             CREATE DATABASE IF NOT EXISTS test
2020-12-14T14:37:57.991905641Z                 encoding='\''utf-8'\'' connection limit=-1
2020-12-14T14:37:57.991917543Z             ;
2020-12-14T14:37:57.991925537Z             GRANT ALL ON DATABASE test TO test;
2020-12-14T14:37:57.991933609Z             GRANT ALL ON DATABASE test TO admin;
2020-12-14T14:37:57.991942499Z             CREATE SCHEDULE test_scheduled_backup
2020-12-14T14:37:57.991951565Z               FOR BACKUP DATABASE test INTO '\''s3://backups/test?AWS_ACCESS_KEY_ID=minioadmin&AWS_SECRET_ACCESS_KEY=minioadmin&AWS_REGION=us-east-1&AWS_ENDPOINT=http://minio.minio:9000'\''
2020-12-14T14:37:57.991964583Z               RECURRING '\''0 * * * *'\''
2020-12-14T14:37:57.991972903Z               FULL BACKUP ALWAYS
2020-12-14T14:37:57.991981115Z               WITH SCHEDULE OPTIONS first_run = '\''now'\''
2020-12-14T14:37:57.991989379Z             ;
2020-12-14T14:37:57.991997224Z             CREATE DATABASE IF NOT EXISTS test2
2020-12-14T14:37:57.992005661Z             ;
2020-12-14T14:37:57.992013970Z             GRANT ALL ON DATABASE test2 TO test;
2020-12-14T14:37:57.992022378Z 
2020-12-14T14:37:57.992030223Z           SELECT '\''Provisioning completed successfully'\'';
2020-12-14T14:37:57.992040473Z         '
2020-12-14T14:37:59.674682073Z ERROR: backups already created in s3://backups/test?AWS_ACCESS_KEY_ID=minioadmin&AWS_ENDPOINT=http%3A%2F%2Fminio.minio%3A9000&AWS_REGION=us-east-1&AWS_SECRET_ACCESS_KEY=minioadmin; to ignore existing backups, the schedule can be created with the 'ignore_existing_backups' option
2020-12-14T14:37:59.674751097Z Failed running "sql"
```

CRDB suggests to use `ignore_existing_backups` or can i ignore some error types/skip them, how?

cockroachdb/templates/job.init.yaml
```diff
@@ -95,29 +101,138 @@ spec:
...
+              }
+
+              initCluster;
+            {{- end }}
+
+            {{- if $isDatabaseInitEnabled }}
+              initDatabases() {
+                while true; do
+                  local output=$(
+                    set -x;
```
sergeyshaykhullin, line 144: Should i remove set -x to not to report user passwords and just use return core to identify that everything is ok? Is it good for troubleshooting?

cockroachdb/values.yaml
```diff
@@ -341,6 +342,30 @@ init:
     #   cpu: "10m"
     #   memory: "128Mi"
 
+  # TODO: dbms? cluster? provisioning? initialization? configuration?
```
sergeyshaykhullin, line 345: TODO: Disable provisioning by default, add comments + examples

cockroachdb/templates/job.init.yaml
```diff
@@ -86,6 +91,7 @@ spec:
         - name: cluster-init
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          # TODO: COMMENT
```
sergeyshaykhullin, line 94: Do not forget to update a comment!

johnrk:
Thanks, @sergeyshaykhullin ! Are you aware of our [open-source Kubernetes Operator](https://github.com/cockroachdb/cockroach-operator)?

Chili-Man:
@johnrk will there be a Helm chart to deploy the kubernetes operator ?

sergeyshaykhullin:
@johnrk Yes, i know it is in alpha now and has no ServiceMonitor, Ingress and provisioning support.
Operators mostly better, because they are Going, not Helming =)

But this one has no required features =(

Same thing with RabbitMQ operator https://github.com/rabbitmq/cluster-operator, but bitnami helm chart has a lot of required features https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq

johnrk:
@sergeyshaykhullin , thanks for the feedback. We will get there with the Operator eventually.

sergeyshaykhullin:
@DuskEagle @pbardea @nathanstilwell @RaduBerinde @johnrk

sergeyshaykhullin:
I have an idea to support not only organization and license, but all cluster settings:
https://www.cockroachlabs.com/docs/stable/cluster-settings.html
https://www.cockroachlabs.com/docs/stable/set-cluster-setting.html

```yaml
init:
  database:
    enabled: true
    clusterSettings:
      cluster.organization: "'FooCorp - Local Testing'"
      enterprise.license: "'xxxxx'"
      sql.metrics.statement_details.enabled: true
      # ...
```

sergeyshaykhullin:
With quoted setting all(duration, boolean, float, integer, enumeration, string, byte size) is working fine

```sql
root@:26257/defaultdb> SET CLUSTER SETTING sql.metrics.statement_details.enabled = "true";
SET CLUSTER SETTING

Time: 338ms total (execution 338ms / network 0ms)

root@:26257/defaultdb> SET CLUSTER SETTING sql.stats.automatic_collection.fraction_stale_rows = '0.2';
SET CLUSTER SETTING

Time: 742ms total (execution 739ms / network 3ms)

root@:26257/defaultdb> SET CLUSTER SETTING sql.temp_object_cleaner.cleanup_interval = '30m0s';
SET CLUSTER SETTING

Time: 244ms total (execution 240ms / network 4ms)

root@:26257/defaultdb> SET CLUSTER SETTING sql.stats.automatic_collection.min_stale_rows = '500';
SET CLUSTER SETTING

Time: 364ms total (execution 349ms / network 15ms)

root@:26257/defaultdb> SET CLUSTER SETTING sql.defaults.serial_normalization = 'rowid';
SET CLUSTER SETTING

Time: 457ms total (execution 456ms / network 0ms)

root@:26257/defaultdb> SET CLUSTER SETTING sql.defaults.results_buffer.size = '16 KiB';
SET CLUSTER SETTING

Time: 166ms total (execution 163ms / network 3ms)

root@:26257/defaultdb>
```

**BUT** setting DEFAULT is not =( I think that it is fine to not to allow resetting settings in helm chart. I got some problems with range end empty values, so skip this for now

```sql
root@:26257/defaultdb> SET CLUSTER SETTING sql.metrics.statement_details.enabled = 'DEFAULT';
ERROR: could not parse "DEFAULT" as type bool: invalid bool value
SQLSTATE: 22P02
```

sergeyshaykhullin:
values
```yaml
init:
  # ...
  database:
    enabled: true
    # https://www.cockroachlabs.com/docs/stable/cluster-settings.html
    clusterSettings:
      cluster.organization: organization
      enterprise.license: license
    users:
    - name: usr
      password: pwd
    #   # https://www.cockroachlabs.com/docs/stable/create-user.html#parameters
    #   options: [LOGIN]
    databases:
    - name: db
    #   # https://www.cockroachlabs.com/docs/stable/create-database.html#parameters
    #   options: [encoding='utf-8']
    #   owners: []
    #   backup:
    #     into: s3://
    #     # Enterprise-only option (revision_history)
    #     # https://www.cockroachlabs.com/docs/stable/create-schedule-for-backup.html#backup-options
    #     options: [revision_history]
    #     recurring: '@always'
    #     # Enterprise-only feature. Remove this value to use `FULL BACKUP ALWAYS`
    #     fullBackup: '@daily'
    #     schedule:
    #       # https://www.cockroachlabs.com/docs/stable/create-schedule-for-backup.html#schedule-options
    #       options: [first_run = 'now']
```

secrets
```yaml
# Source: cockroachdb/templates/secrets.init.yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-release-cockroachdb-init
type: Opaque
stringData:
  usr-password: pwd
  cluster-organization-cluster-setting: organization
  enterprise-license-cluster-setting: license
```

job
```yaml
# Source: cockroachdb/templates/job.init.yaml
kind: Job
apiVersion: batch/v1
metadata:
  name: test-release-cockroachdb-init
  namespace: "test-namespace"
  labels:
    helm.sh/chart: cockroachdb-5.1.0
    app.kubernetes.io/name: cockroachdb
    app.kubernetes.io/instance: "test-release"
    app.kubernetes.io/managed-by: "Helm"
    app.kubernetes.io/component: init
  annotations:
    helm.sh/hook-delete-policy: before-hook-creation
spec:
  template:
    metadata:
      labels:
        app.kubernetes.io/name: cockroachdb
        app.kubernetes.io/instance: "test-release"
        app.kubernetes.io/component: init
    spec:
      restartPolicy: OnFailure
      terminationGracePeriodSeconds: 0
      containers:
        - name: cluster-init
          image: "cockroachdb/cockroach:v20.2.3"
          imagePullPolicy: "IfNotPresent"
          # Run the command in an `while true` loop because this Job is bound
          # to come up before the CockroachDB Pods (due to the time needed to
          # get PersistentVolumes attached to Nodes), and sleeping 5 seconds
          # between attempts is much better than letting the Pod fail when
          # the init command does and waiting out Kubernetes' non-configurable
          # exponential back-off for Pod restarts.
          # Command completes either when cluster initialization succeeds,
          # or when cluster has been initialized already.
          command:
          - /bin/bash
          - -c
          - >-
              initCluster() {
                while true; do
                  local output=$(
                    set -x;

                    /cockroach/cockroach init \
                      --insecure \
                      --host=test-release-cockroachdb-0.test-release-cockroachdb:26257
                  2>&1);

                  local returnCode="$?";
                  echo $output;

                  if [[ "$returnCode" == "0" || "$output" == *"cluster has already been initialized"* ]]
                    then break;
                  fi

                  sleep 5;
                done
              }

              initCluster;
              initDatabases() {
                while true; do
                  local output=$(
                    set -x;

                    /cockroach/cockroach sql \
                      --insecure \
                      --host=test-release-cockroachdb-0.test-release-cockroachdb:26257 \
                      --execute="
                          SET CLUSTER SETTING cluster.organization = '$cluster_organization_CLUSTER_SETTING';
                          SET CLUSTER SETTING enterprise.license = '$enterprise_license_CLUSTER_SETTING';
                          CREATE USER IF NOT EXISTS usr WITH
                            PASSWORD '$usr_PASSWORD'
                            
                          ;
                          CREATE DATABASE IF NOT EXISTS db
                          ;

                          SELECT 'Provisioning completed successfully';
                      "
                    2>&1);

                  local returnCode="$?";

                  if [[ "$returnCode" == "0" || "$output" == *"Provisioning completed successfully"* ]]
                    then break;
                  fi

                  sleep 5;
                done
              }

              initDatabases;
          env:
          - name: usr_PASSWORD
            valueFrom:
              secretKeyRef:
                name: test-release-cockroachdb-init
                key: usr-password
          - name: cluster_organization_CLUSTER_SETTING
            valueFrom:
              secretKeyRef:
                name: test-release-cockroachdb-init
                key: cluster-organization-cluster-setting
          - name: enterprise_license_CLUSTER_SETTING
            valueFrom:
              secretKeyRef:
                name: test-release-cockroachdb-init
                key: enterprise-license-cluster-setting
```

sergeyshaykhullin:
Hmm...

Some CI problems i guess

```
[13:21:18][Step 1/1] Adding stable repo with URL: https://kubernetes-charts.storage.googleapis.com 
[13:21:19][Step 1/1] Error: error initializing: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden
[13:21:19][Step 1/1] Error linting charts: Error initializing Helm: Error waiting for process: exit status 1
```

sergeyshaykhullin:
@johnrk Can you schedule someone for a review, please?

keith-mcclellan:
My big issue here is the secrets (namely the users password and license key) are going to be available in plain-text to anyone with appropriate permissions to the Kubernetes cluster.  I would rather we inject those values into secrets so at least we're storing them securely.

cockroachdb/templates/job.init.yaml
keith-mcclellan: set -x is going to print the secrets to the log in plain text so this needs to be removed.  We actually need to go the other way and explicitly route the commands where we're injecting secrets to >/dev/null so we never print those values to the console or the logs

keith-mcclellan (changes requested):
The issue I have here is that the secrets (license key and user password) will get printed to the job log console, so they won't be secure.  It's okay to have a debug flag for troubleshooting that will print this to the log for debug, but the default needs to route stdout to /dev/null so we don't print it anywhere.

sergeyshaykhullin:
crdb

sergeyshaykhullin:
@keith-mcclellan @johnrk Can we try one more time?)

keith-mcclellan:
@chrisseto @DuskEagle @joshimhoff Would one of you please give this a final review?  This is a big enough change that I think it makes sense to have a second opinion here. Thanks!!

PaulFarver:
This would be a great feature!

I realize that I'm sort of just "popping" in from the side a bit, but I was about to submit a similar PR, when I noticed this one, and wanted to just give my two cents here.

- In order to simplify the init job spec, you could just put all the templated sql in a secret, and mount that as a file and run it with sql. That way you wouldn't have to add all user passwords and cluster settings in environment variables. Alternatively you could use `envFrom` to avoid having to loop through all keys in the secrets.

- We started to build a similar solution internally, but had to switch over to just running raw sql, because the sql templating started to get really messy when we needed to update passwords and create more advanced permissions. It would be nice to still have the option to run what you want. Would that be possible?

sergeyshaykhullin:
@PaulFarver 

> In order to simplify the init job spec, you could just put all the templated sql in a secret

I think it might be a case, but my point is to store in secrets only secret info. I will minimize job manifest, but make secret unclear.

Now i can get secret key:value pairs inside to analyze thats and why's and job command with clear env vars naming.
If i will put all info inside secret and template sql text with helm - i will lose info where is value from

Example:

```yaml
users: [{ username: user1, password: same }, { username: user2, password: same }, { username: same, password: same }]
databases: [{ name: same }]
```

> Had to switch over to just running raw sql

I thought to generalize this pr to just sql runner with sql templating on top of it(init, provision, raw), but started with concrete requirement - add ability to pre-configure cluster with helm. This issue has some clarifications https://github.com/cockroachdb/helm-charts/issues/7 that i wanted to support

If you want to refactor something here or implement a better way - please, welcome!)

cockroachdb/templates/job.init.yaml
```diff
@@ -1,4 +1,6 @@
-{{- if and (eq (len .Values.conf.join) 0) (not (index .Values.conf `single-node`)) }}
+{{ $isClusterInitEnabled := and (eq (len .Values.conf.join) 0) (not (index .Values.conf `single-node`)) }}
+{{ $isClusterProvisioningEnabled := .Values.init.provisioning.enabled }}
+{{- if or $isClusterInitEnabled $isClusterProvisioningEnabled }}
```
udnay, line 3: Is it ok/possible for both of these values to be true? Do you ever want to run the provisioning if `$isClusterInitEnabled` is false?

cockroachdb/templates/job.init.yaml
```diff
@@ -119,29 +121,140 @@ spec:
...
+
+                  if [[ "$exitCode" == "0" ]]
+                    then break;
+                  fi
+
+                  sleep 5;
+                done
+              }
+
+              provisionCluster;
```
udnay, line 236: Can this be wrapped in a timeout so that it eventually breaks the loop?

cockroachdb/templates/job.init.yaml
sergeyshaykhullin: Yes, for single node with preconfigured databases

cockroachdb/templates/job.init.yaml
sergeyshaykhullin: Hmm... Can you please provide an example?

udnay: What happens if `/cockroach/cockroach sql` exits with a non-zero value, when does the init container exit the while loop?

I was thinking of the [timeout command](https://man7.org/linux/man-pages/man1/timeout.1.html) so something like `timeout 30 provisionCluster;` is there something else that will cause the init container to stop after a certain amount of time?

sergeyshaykhullin: If sql command exit with non-zero code you will get infinite loop with 5 seconds interval. You should get logs and try to find root cause of non completed job. Same behaviour as before, now init job is doing the same

What will happen if sql script is valid, but not ran to completion in 30 seconds? I think that completed job without any guarantee of applied sql is unclear and mostly harmful

Also timeout configuration can be very tricky

cockroachdb/templates/job.init.yaml
```diff
@@ -119,29 +121,140 @@ spec:
...
+                      --cluster-name={{.}} \
+                      {{- end }}
+                      --host={{ template "cockroachdb.fullname" . }}-0.{{ template "cockroachdb.fullname" . -}}
+                              :{{ .Values.service.ports.grpc.internal.port | int64 }}
+                  2>&1);
+
+                  local exitCode="$?";
+                  echo $output;
+
+                  if [[ "$exitCode" == "0" || "$output" == *"cluster has already been initialized"* ]]
```
udnay, line 149: @keith-mcclellan is there a way to check if a cluster is up vs grep'ing for a string?

cockroachdb/templates/job.init.yaml
udnay: And the sql script is idempotent? The 30 seconds was arbitrary but you're right that half applied sql statements are more likely to put the DB into a bad state. If the script is continuously failing though won't the DB be in a bad state already potentially? @keith-mcclellan are you ok with this workflow?

sergeyshaykhullin: Yes, all except backups are idempontent https://github.com/cockroachdb/cockroach/issues/57892

udnay: Ok thanks for the clarification

cockroachdb/templates/job.init.yaml
DuskEagle: I'm not aware of one. This is at least the same as what we were doing before this PR, so I'm fine with continuing to use it.

cockroachdb/templates/job.init.yaml
```diff
@@ -119,29 +121,140 @@ spec:
...
+                        {{- end }}
+
+                        {{- if and $database.backup.schedule $database.backup.schedule.options }}
+                          WITH SCHEDULE OPTIONS {{ join "," $database.backup.schedule.options }}
+                        {{- end }}
+                        ;
+                      {{- end }}
+                      {{- end }}
+
+                        SELECT 'Provisioning completed successfully';
```
DuskEagle, line 222: Is there a reason for logging this using a SQL statement?

DuskEagle (commented):
Looks good, just a couple small questions.

cockroachdb/templates/job.init.yaml
sergeyshaykhullin: @DuskEagle To be honest, I don't remember why. I think it can be logged after the loop

cockroachdb/templates/job.init.yaml
sergeyshaykhullin: I think i can clarify usage and rename to $isDatabaseProvisioningEnabled

DuskEagle (approved):
LGTM

sergeyshaykhullin:
@DuskEagle @keith-mcclellan Does this pr can be merged?

adityamaru:
fwiw we should have https://github.com/cockroachdb/cockroach/issues/57892 prioritized soon!

sergeyshaykhullin:
❤️

sergeyshaykhullin:
I think crdb will ship idemponent schedule creation in next release https://github.com/cockroachdb/cockroach/pull/69152

This pr is done, waiting for review/merge

udnay:
Can you add some test cases to validate the template changes please?

sergeyshaykhullin:
@udnay Done

alborotogarcia (changes requested):
add password within parenthesis and remove simple quotes

cockroachdb/templates/job.init.yaml
```diff
@@ -86,29 +88,140 @@ spec:
...
+                            :{{ .Values.service.ports.grpc.internal.port | int64 }} \
+                    --execute="
+                      {{- range $clusterSetting, $clusterSettingValue := .Values.init.provisioning.clusterSettings }}
+                        SET CLUSTER SETTING {{ $clusterSetting }} = '${{ $clusterSetting | replace "." "_" }}_CLUSTER_SETTING';
+                      {{- end }}
+
+                      {{- range $user := .Values.init.provisioning.users }}
+                        CREATE USER IF NOT EXISTS {{ $user.name }} WITH
+                        {{- if $user.password }}
+                          PASSWORD '${{ $user.name }}_PASSWORD'
```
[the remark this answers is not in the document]
sergeyshaykhullin: Can you please add more datails? Can you share username(or same structured username)?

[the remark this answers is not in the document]
sergeyshaykhullin: Would be asome if you can share your values.yaml section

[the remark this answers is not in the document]
alborotogarcia: there you go..
diff --git a/cockroachdb/values.yaml b/cockroachdb/values.yaml
index 9783821..05f64c1 100644
--- a/cockroachdb/values.yaml
+++ b/cockroachdb/values.yaml
@@ -1,9 +1,9 @@
 image:
-  tag: v21.1.7
+  tag: 21.1.8
   pullPolicy: IfNotPresent
-  credentials: {}
-    # registry: docker.io
+  credentials:
+    registry: docker.io
     # username: john_doe
     # password: changeme
 
@@ -336,7 +336,7 @@ storage:
   persistentVolume:
     enabled: true
 
-    size: 100Gi
+    size: 5Gi
 
     # If defined, then `storageClassName: <storageClass>`.
     # If set to "-", then `storageClassName: ""`, which disables dynamic
@@ -344,7 +344,7 @@ storage:
     # If undefined or empty (default), then no `storageClassName` spec is set,
     # so the default provisioner will be chosen (gp2 on AWS, standard on
     # GKE, AWS & OpenStack).
-    storageClass: ""
+    storageClass: "longhorn"
 
     # Additional labels to apply to the created PersistentVolumeClaims.
     labels: {}
@@ -388,21 +388,34 @@ init:
     #   memory: "128Mi"
 
   provisioning:
-    enabled: false
+    enabled: true
     # https://www.cockroachlabs.com/docs/stable/cluster-settings.html
-    clusterSettings:
+    clusterSettings: {}
       # cluster.organization: "'FooCorp - Local Testing'"
       # enterprise.license: "'xxxxx'"
-    users: []
-    # - name:
-    #   password:
+    users:
+    - name: alboroto
+      password: '!pass123'
+      options: [LOGIN, CREATEROLE]
+    - name: mlflow
+      password: mlflow
+      options: [LOGIN]
+    - name: superset
+      password: superset
+      options: [LOGIN]
     #   # https://www.cockroachlabs.com/docs/stable/create-user.html#parameters
     #   options: [LOGIN]
-    databases: []
-    # - name:
+    databases:
+    - name: mlflow
     #   # https://www.cockroachlabs.com/docs/stable/create-database.html#parameters
     #   options: [encoding='utf-8']
-    #   owners: []
+      owners: [mlflow]
+    - name: testdb
+      owners: [alboroto]
+    - name: testdb
+      owners: [alboroto]
+    - name: superset
+      owners: [superset]
     #   # Backup schedules are not idemponent for now and will fail on next run
     #   # https://github.com/cockroachdb/cockroach/issues/57892
     #   backup:
 networkPolicy:
   enabled: false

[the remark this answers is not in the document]
sergeyshaykhullin: Seems weird)

image

Does all your users without password or one starting from `!`?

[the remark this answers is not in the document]
sergeyshaykhullin: I will test real deployment with this values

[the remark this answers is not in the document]
sergeyshaykhullin: image

image

Missed namespace and quotation. It will work only in default namespace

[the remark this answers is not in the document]
sergeyshaykhullin:
```
++ /cockroach/cockroach init --certs-dir=/cockroach-certs/ --host=test-release-cockroachdb-0.test-release-cockroachdb:26257
warning: node not ready to perform cluster initialization: initial connection heartbeat failed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing dial tcp 10.42.2.103:26257: connect: connection refused" (retrying)
Cluster successfully initialized
GRANT

          CREATE USER IF NOT EXISTS alboroto WITH
            PASSWORD '!pass123'
            LOGIN CREATEROLE
          ;
          CREATE USER IF NOT EXISTS mlflow WITH
            PASSWORD 'mlflow'
            LOGIN
          ;
          CREATE USER IF NOT EXISTS superset WITH
            PASSWORD 'superset'
            LOGIN
          ;
          CREATE DATABASE IF NOT EXISTS mlflow
          ;
          GRANT ALL ON DATABASE mlflow TO mlflow;
          CREATE DATABASE IF NOT EXISTS testdb
          ;
          GRANT ALL ON DATABASE testdb TO alboroto;
          CREATE DATABASE IF NOT EXISTS testdb
          ;
          GRANT ALL ON DATABASE testdb TO alboroto;
          CREATE DATABASE IF NOT EXISTS superset
          ;
          GRANT ALL ON DATABASE superset TO superset;
      
Provisioning completed successfully
```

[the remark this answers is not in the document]
alborotogarcia: Couldn't login with any.. I guess that special characters breaks the user creation loop.. Btw is there a way to set root password or guess it afterwards accessing to nodes?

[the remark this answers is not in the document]
sergeyshaykhullin: @alborotogarcia Does `$({{ $user.name }}_PASSWORD)` gives you correct passwords? Did you changed something else? Secret was in default namespace, so you can't mount its values to init job from `another-namespace`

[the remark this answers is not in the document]
sergeyshaykhullin: Does init job logs something?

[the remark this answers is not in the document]
sergeyshaykhullin: Also you can specify passwords only for secure clusters. Which are you running?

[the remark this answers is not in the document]
sergeyshaykhullin: You can't change root(and any) password, only create new user for now. Because ALTER USER is not supported

https://www.cockroachlabs.com/docs/stable/authorization.html#root-user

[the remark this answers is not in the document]
alborotogarcia: Yes you were right, previously I created them as sealed secrets to the namespace, as I read from this PR that it should be taken from vault secrets, and as it was giving me errors I left them..
```apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  creationTimestamp: null
  name: crdb-cockroachdb-init
  namespace: storage
spec:
  encryptedData:
    alboroto-password: AgB2TYo2b3Jd2Wwah31vxnOMoVrv6fZeAmlaEjWOyScio72YV0teLdYNaHoxDQc2h96kDtBbsFAtlPcGsv5IRWX49ebkSHaAkIKn+RtEANbhgiMnujImb7Vnxmw7WZy/tO5qFnkaHKDceAR5/zJEkcFM4Jt+7yLrCLsihvTO3BhgDgBdfBxPscS0bHTP6gULGcawh2AHo7/wcascqcVWeMMHCVeEvgqJkF0DkkA07zxGMsQS/FmxXcKVl1H5BK5TeKpTUj/pFULNdNPkkG3tvKUG0OBYKRk6GjSNcKk5XKd+KrewVh2JjwVicolTXjbzY0Y5r6MK6H7KcQaGc1Kr8iFRhXyuZbp8zHvU8mcdzaU18MUKl8CSSBLud9BKr0+rGpiImzUtIMXAZJKVSaxsT+uR8MjqinxKRHOIgeZ1T9DiRZx962dqHgdArbXlKDEF1ekfahtL7j6doj8nujuCZvM+CMxOWXInplsL2LeWEgKB9hVTOOUWkqkhywGteJ24+W4/Kkuxv8BIy2/QuR9LuAlBV5bGpS9fl+T7eIe7Q0gI5D81bsKbCPASSe8SLoewUAIJM/iFyV0BpOcydfa30UzhXrkOGabHeT5PHuL65GxXlXUuPAq9k3NcUoCsZy6pPoWOMaBOmAFS95PEEQkY4RGEyQ5kKU34c1poNyHICF+IcfqKMOpxNQcXgGyMCu3daH5wrCwMD6RiQaOt
    mlflow-password: AgBDORP4PqhDO17fPVeyXA1IH1ExiOoF8x6G4FA34DjwP0yOABwK4wycSpx7koTysbCldLVKta3P1cH6aqFJ7IyQ+bFxRzaoK1kxfCCeakIckTZ9OoDOQAaY71anZYgb9PE54ADXhDxlR4MiTkwRwSwcAYAlERHXj1tjgnNuCz+qc1cg9dHL5zfphNdEgf69fyBq/cGeRn/m/gvr8EdS3u2KwQ98Z4htBQs1Sazew9pJPjPUkPv/CNXoCwH+SX+c3N9qD3vdVoBfUMZCFZI7qRa+NqxQHzXqXw3nNQaW8MBl+enK9NmWJQHShFMxEAMRenfurC65U+goXhPV8zL6iS6Mr2I45MInzeQy8vuUTHT60GQIZ8aILIGF14eVs6hEuMA6GZRppOapQCyzIKpM+3R4m18j+0E0EaHe0/7Hb2nGPqUU0mXPVW3/rckVCChlJNVq2UCpt0BEKHi5bzLubYioHiYWwoHZx83eTKAgORvRJxjR5EUndBLajeXkp8WRgerCwwfvOfZJPcHQhn/kwSsxQYaINegK4aXtIgWeOnyS8Pmh+k0uMOMpmWv1DvbWd1SZl0NAVC3nBYescTRxt1VrSBIWwYbVTmmfTcgXcgrbiPA+uZrncjj7Jhpoogc3TMU1f8Z2lrBngfG8u2zyBmFhX67wTrDJzIZOEyIO/f2dXJaGh4g+wqHrmxRsjYaQABNFunD8h5k=
    postgres-password: AgBdyPSwRDU5IEHinRNvBQaK10adfKPt5xmR2epxYM6w3YU9l2pChIgbwCp4J9tW/2AQkpEgvHM+XWOyv/syPRjoLa11lCcHhKLrB67fEtCb470WxJP09IJTv908K+/qo8RLi0SURvdjN+HRT39P+GI5hZre+36/tqAARXCvv+z5QMfuFIPZaMkz6giVs6+NrmsWhYi6JvIxwL62cujaOS+5wq7VIJ5D049t76DfTdUifeTBJnChEY/IXkc61BRQhwzSrM0JOpzXNh3rc56bR8Sk56ksixbYPXNKA6PKI0H9dCFjdNt5bFSkvIgwIFyPLNMc7655EuQO43TOT/VHvAk6mC7tJfVyHiiWoLNmUFwqaRvZU7mssM69qiz6pC2M/lZ82BVDfa0hfgMitaUPVh297JkEW43iHOBv4QIN6N7nx/xSJ5UttwUf2U4/AoFZJSEDmq8RsaAKqNrCPrHKha55yOlEjMVvGx70V5ybRfzfhqQjiO+kFrnQWJbtHoQLqsdUPmCXC0BHolrqPF9GhL75nN8PGkQEDejsKdYedCyysmNvwDd2tfY5NThEeYjxJKdesg3NfCofWNDWZcDzJBJpU9Vyt2BwjiJhqXXg2RR5iKZwO1lDwisaSqbuDLtiCqurn8PcND6R/Sy4LNGmDrSaEtyjtMmqjvpcjvTj5gPa/4xEY+78JIRs2hSO7lzBA8DNmHFDYr26Ng==
    superset-password: AgB6NIpMPUMgcCUMIK7xgZvE9555xZAr485FYA9TYV77r64F+yIDC2DwhQGEReuhoyBkYHqBJHwGaBrVCoYJ1z/g/evUeG49cl/N5flGF/3YXAbrgP7VxJesxxZHIAQ6cRrRkQgtDtee0rjGFAj20AbhrjKW1BU9n16AxT6xvXIGS2jjwrl5ZG6+ziGH4+wmD2jSoZ11Bru/rmc7gNTWRb2RygyimXx7RdUQKni01A+DVelCaY1AESLeZoJNKGrmXTtH1CNg3+CXxhc/OqEkkeDk14+Bs+Lr+znr736xbwbqyBVTH5Nw7XRu0iK/utquqGL8iUt6yDIcUyyMv2N7TQtRq0R1N+GTXuEezzA5s+k/KYHXI553hoABCEzwbXlgeT03GLcr/cVZ0G0xvdceE1mGE6xDOHQkIQasT0A52LLH5IAHPV30Yhi59NOVAI5UavbgGe/yA5DHYrXl8lxowsD54ctm0QcX5BF0cQA54wPQRVH99M84PDs0trgSpV/vFkWYk7bOLI3cy2S88f4dodXrHFqtnKH3Izf8xj06iu52CEzhvM1Rb2ZeTvkTLK9g5qHgOF1nqdbOX/HgPwmB3cWfQ7IRq7sIKe/XZ3tsABdWnpVUVZ7HLp8UYD8kavhcCxq6gFheUG4yBTEAtX2ZI2c2up3XexLMiCrTEbZ5ORxr8yeQ16+SEh6icsZ5jxBMvGJvf/AwovwDWw==
  template:
    data: null
    metadata:
      creationTimestamp: null
      name: crdb-cockroachdb-init
      namespace: storage
    type: Opaque
```


```(base) [alvaro@stealth cockroachdb]$  git diff templates/
diff --git a/cockroachdb/templates/job.init.yaml b/cockroachdb/templates/job.init.yaml
index 7c7c758..84b9dc2 100644
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -146,7 +146,7 @@ spec:
                       {{- range $user := .Values.init.provisioning.users }}
                         CREATE USER IF NOT EXISTS {{ $user.name }} WITH
                         {{- if $user.password }}
-                          PASSWORD '${{ $user.name }}_PASSWORD'
+                          PASSWORD $({{ $user.name }}_PASSWORD)
                         {{- else }}
                           PASSWORD null
                         {{- end }}
diff --git a/cockroachdb/templates/statefulset.yaml b/cockroachdb/templates/statefulset.yaml
index 8572b80..24e5fc4 100644
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -143,8 +143,8 @@ spec:
         - name: db
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
           imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          command: [sh, -c]
           args:
-            - shell
             - -ecx
             # The use of qualified `hostname -f` is crucial:
             # Other nodes aren't able to look up the unqualified hostname.
`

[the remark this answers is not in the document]
Comment (sergeyshaykhullin) [reply cockroachdb/templates/job.init.yaml:149]: Tested with provisioned user

```sh
[root@test-release-cockroachdb-0 cockroach]# cockroach sql --certs-dir cockroach-certs/ --user testUser
#
# Welcome to the CockroachDB SQL shell.
# All statements must be terminated by a semicolon.
# To exit, type: \q.
#
Enter password: 
# Server version: CockroachDB CCL v21.1.7 (x86_64-unknown-linux-gnu, built 2021/08/09 17:55:28, go1.15.14) (same version as client)
# Cluster ID: 320ec220-2f65-4b77-a36b-fcf07bf23f88
#
# Enter \? for a brief introduction.
#
testUser@:26257/defaultdb> select 1;
  ?column?
------------
         1
(1 row)

Time: 1ms total (execution 1ms / network 1ms)

testUser@:26257/defaultdb>
```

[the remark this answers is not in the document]
alborotogarcia: Great @sergeyshaykhullin ! Working flawlessly !

alborotogarcia (approved):

udnay (approved):

udnay:
LGTM, can you just bump the version before I merge?

sergeyshaykhullin:
@udnay Can you please clarify? Version is already bumped from 6.1.2 to 6.2.0. Do you mean major? There is no breaking changes

udnay:
@keith-mcclellan are you comfortable with the `set -x` in here by default, I can't tell if it's being redirected to `dev/null`, I don't think so but I might be missing it.

sergeyshaykhullin:
@udnay set -x is from previous implementation

image

sergeyshaykhullin:
@udnay @keith-mcclellan @DuskEagle Any updates?

udnay:
I think the `set -x` is ok if it's legacy

PR #50, opened 2021-01-21
Update cockroachdb chart to version 20.2.4

pbardea:
Used https://github.com/cockroachdb/helm-charts/commit/e381aa73e6bed05b42b6c7eda9f63398efa37f1a as template.

PR #51, opened 2021-01-22
Upgrade the chart testing version

jlinder:
The default stable chart repository moved and the old testing version pointed
at the old default stable chart repository URL. The upgraded version
fixes that.


jlinder:
@pbardea This should fix the issue you're having in #50 .

pbardea (approved):
Thanks!

jlinder:
Sure thing!

PR #66, opened 2021-03-31
Add store configuration

sergeyshaykhullin:
## Problem

For now CRDB is using (and showing) that he can use full node disk space, but it is not true

I found storage configuration section https://www.cockroachlabs.com/docs/v20.2/cockroach-start.html#storage

And implemented spec expression in helm

## Results
```yaml
store:
    enabled: true

--store=path=cockroach-data,size=100Gi
```

```yaml
store:
    enabled: true
    type: mem
    size: 1Gi

--store=size=1Gi,type=mem
```

```yaml
store:
    enabled: true
    size: 1Gi

--store=size=1Gi,path=cockroach-data
```

```yaml
store:
    enabled: true
    size: 1Gi

--store=size=1Gi,path=cockroach-data
```

```yaml
store:
    enabled: true
    attrs: hdd:7200rpm

--store=path=cockroach-data,size=100Gi,attrs=hdd:7200rpm
```

```yaml
store:
    enabled: true
    path: my-data
    size: 404Gi
    attrs: hdd:7200rpm

--store=size=404Gi,attrs=hdd:7200rpm,path=my-data
```

image

## Problems
- [ ] Do we really need `enabled`? I think it should be enabled for all clusters
- [ ] `dict` ordering, sometimes `path` can swap with someone... Not tested for validity
- [ ] This will not work if `statefulset.args` has it's own `--store`


cockroachdb/values.yaml
```diff
@@ -134,6 +134,16 @@ conf:
   # CockroachDB's port to listen to HTTP requests.
   http-port: 8080
 
+  # CockroachDB's storage configuration https://www.cockroachlabs.com/docs/v20.2/cockroach-start.html#storage
```
DuskEagle, line 137: Maybe mention that this is for supplying values to the optional `--store` field, and mention that this is only needed if `storage.persistentVolume.enabled` == `false`?

cockroachdb/values.yaml
```diff
@@ -134,6 +134,16 @@ conf:
...
+  store:
+    enabled: false
```
DuskEagle, line 139: It's a little strange that `enabled` can be `false`, but that the `store.path` field is still read for the `mountPath`.  Maybe `store.path` should be moved to the top level of this config?

cockroachdb/values.yaml
```diff
@@ -134,6 +134,16 @@ conf:
...
+    # Should be empty or 'mem'
+    type:
+    path: cockroach-data
+    # Required for type=mem. If type and size is empty - storage.persistentVolume.size is used
+    size:
+    # hdd:7200rpm
```
DuskEagle, line 145: This line looks extraneous.

cockroachdb/values.yaml
```diff
@@ -134,6 +134,16 @@ conf:
...
+    attrs:
```
DuskEagle, line 146: Can you add a short description for each of these attributes to the [configuration section of the README.md](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/README.md#configuration)?

cockroachdb/values.yaml
sergeyshaykhullin: No, you can use it with disabled PVs. Because of inmem mode

sergeyshaykhullin: Why? You can use for both persistent and inmem

DuskEagle: I see a reason to use it for in-memory or if using `hostPath`, but I don't really know why you would use it when using persistent volumes. But you can still use it for persistent volumes if you want.

DuskEagle (approved):

cockroachdb/values.yaml
sergeyshaykhullin: @DuskEagle Hmm, my main usecase is for persistent deployments to have a storage limit.

Example:
I have 3 nodes, 100Gb disks
One cluster with 10Gb PVs
CRDB will think he can use all 100Gb because of 100% default storage limit, but it is not true

PR #68, opened 2021-04-09
ensure the init job is created as a helm hook so the chart can be upgraded

### LINKED ISSUE #45: Init job as a helm hook

JorritSalverda:
fixes #45

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 5.0.7
+version: 5.0.8
 appVersion: 20.2.7
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
 icon: https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png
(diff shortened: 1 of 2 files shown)


JorritSalverda:
Hi @JuanLeon1 @DuskEagle any chance this can be approved, merged and released in the next version?

DuskEagle:
:LGTM: if you could update the `Chart.yaml` version as mentioned in [Contributing.md](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/CONTRIBUTING.md). Thank you!

DuskEagle (approved):

jonasbadstuebner:
Are you sure this works on first install as well?
Because it will not create the hook necessary for init, until _after_ the installation would complete, which it doesn't if you handle it like this.

PR #60, opened 2021-03-17
set correct api version and backend structure for kubernetes 1.19+

JorritSalverda:
Fixes https://github.com/cockroachdb/helm-charts/issues/59

PR #74, opened 2021-04-13
fix use of capabilities for ingress inside loop

JorritSalverda:


JorritSalverda:
Fixes regression caused by https://github.com/cockroachdb/helm-charts/pull/60

JorritSalverda:
@JuanLeon1 better to merge this one so the helm chart works again for ingress. Again, sorry for breaking things.

cockroachdb/templates/ingress.yaml
```diff
@@ -40,11 +40,11 @@ spec:
         paths:
   {{- range $path := $paths }}
           - path: {{ $path }}
-            {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+            {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion -}}
```
keith-mcclellan, line 43: can you please move this to Version instead of GitVersion? like `$.Capabilities.KubeVersion.Version`

cockroachdb/templates/ingress.yaml
```diff
@@ -40,11 +40,11 @@ spec:
...
             pathType: Prefix
             {{- end }}
             backend:
-              {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion -}}
```
keith-mcclellan, line 47: same

cockroachdb/templates/ingress.yaml
```diff
@@ -60,11 +60,11 @@ spec:
         paths:
   {{- range $path := $paths }}
           - path: {{ $path }}
-            {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+            {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion -}}
```
keith-mcclellan, line 63: same

cockroachdb/templates/ingress.yaml
```diff
@@ -60,11 +60,11 @@ spec:
...
             pathType: Prefix
             {{- end }}
             backend:
-              {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
+              {{- if semverCompare ">=1.19-0" $.Capabilities.KubeVersion.GitVersion -}}
```
keith-mcclellan, line 67: same

keith-mcclellan (changes requested):

cockroachdb/templates/ingress.yaml
JorritSalverda: I guess even better if I use:
```
{{ if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
```

So far I just copied what I saw elsewhere in the chart, but will do an update.

keith-mcclellan: We just started making this change - thanks @JorritSalverda !!

keith-mcclellan (approved):
LGTM

keith-mcclellan:
@JorritSalverda could you please rebase so we can merge? ty!

JorritSalverda:
@keith-mcclellan I've merged most recent changes from master. Not sure how to rebase things so it gets a cleaner commit history though.

DuskEagle:
I will squash and merge all these commits into a single commit, so the git history looks clean.

PR #80, opened 2021-04-19
RFC: Certificate Management in cockroachDB Helm chart

prafull01:
This RFC proposes a method of deploying the CockroachDB Helm chart in a secure mode by generating certificates without using Kubernetes CA.  
The new method allows users to specify their own CA or if not provided will generate a self-generated CA.  
It handles the creation of Node and Client certificates using the CA.  
It allows rotation of CA, Node, and Client certificates.  
It allows specifying a duration for the generated certificates

### DIFF
--- a/docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
+++ b/docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
@@ -4,7 +4,7 @@ Status: Draft
  
 Start Date: 19-04-2021
  
-Authors: @prafull01, @abhisek-dwivedi, @madhurnawandar
+Authors: @prafull01, @abhishekdwivedi3060, @madhurnawandar
  
 # Summary
  


docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+   if yes then it will renew the certificates.
+  
+  * <b>The cronjob will use the same `pre-install` job image for certificate rotations. The `pre-install` job image binary will have an argument `--rotate` for handling certificate rotation.</b>
+ 
+* The Stateful pod will be changed to only run `copy-certs` initContainer to copy the certificates from Node secret to emptyDir volume.
+  The rest of the main DB container flow will remain the same.
+* The post-install job will be changed to only run `copy-certs` initContainer to copy the certificates from Client secret to emptyDir volume.
+  The rest of the main cluster-init container flow will remain the same.
+ 
+`TODO: Need to identify how to generate the SIGHUP signal in all the Nodes for certificate renewal`
```
DuskEagle, line 139: One way that we've done this is to install a sidecar container in the CockroachDB pod that listens for changes to the certificate files and SIGHUPs the CRDB process if one is detected. We could introduce this functionality to our Helm chart.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+   `tls.certs.generate.clientCertDuration` and `tls.certs.generate.nodeCertDuration`.
+ 
+ 
+## Implementation Details
+ 
+### Helm Components:
+ 
+When `tls.certs.generate.enabled` is set to `true`, the following components are created for certificate generation and rotation:
+ 1. Certificate Management Service as a `pre-install` job.
+ 2. ServiceAccount, for `pre-install` job. (deleted after pre-install hook succeeds)
```
DuskEagle, line 80: Does the service account need to stick around to support certificate rotation?

DuskEagle (commented):
This looks like a really solid proposal to me! I've requested a few additional eyes from other people I think might be interested in this.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+* In addition, Client secret and Node secret will be patched with annotation `needsRegeneration: True`, which specifies that their certificates
+  need to be regenerated in the next cronjob run. This is done in accordance with the suggestion in CockroachDB [doc](https://www.cockroachlabs.com/docs/v20.2/rotate-certificates.html#why-rotate-ca-certificates-in-advance)
+ 
+#### Check-node-cert-for-regeneration
+* Follow checks from [Check cert for regeneration](#check-cert-for-regeneration)
+* In addition, check if annotation `needsRegeneration: True` exists, if yes return True (This will be case when CA certificate has rotated, but Client and node certs are still to be recreated)
+ 
+#### Check-cert-for-regeneration
+* Check if the certificate secret has annotation `managed-by: cockroachdb`, if not return False
+* Check if the certificate is expiring in the next 2 months using the values from annotations `creationTime` and `duration` on secret, if expiring return True, else False
```
udnay, line 200: Lets make the window where expiring is true configurable as well. `expiryWindow` or something

udnay (commented):
This looks good. Is there some way to fail "loudly" if a cert rotation isn't succeeding is there a way to indicate this or configure it in such a way that the failure isn't just spat out to the logs?

keith-mcclellan (changes requested):
Just a couple of gaps, but in general I'm happy with the scope here.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
prafull01: The need of pre-install hook's service account is only for running the pre-install job to create certificates. This service account doesn't need to stick around because once the helm release has been successfully deployed, we can use crdb service account for running the cronjob later and use that to rotate certificates.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
prafull01: The expiryWindow configurable is not straight forward becuase it leads to many edge cases of certificate rotation. For example:

Lets take three types of certificate and their duration:

| Client          |            Node            |              CA     |
|-----------------|---------------------------|--------------------|
|1 Year          |           2 Years         |           3 Year   |

Expiry window is 2 months (lets say provided by user)

1. The client certificate will rotate with first cron on 10th Month and again with second cron on 20th Month. 
2. During the second cron client certificate will be renewed but as the node certificate is going to expire before next cron, it will try to rotate node certificate on the 20th month. But if this node certificate is rotated with 2 year again it will outlive CA by 1 year, which is not valid.  

We have to device a method/formula with certain constraints to properly handle cert rotation.

One formula/approach could be:
We can have two crons having different schedules, one handle the CA rotation and other handles the client/node certificate rotation.

1. The CA cron schedule should be :  `Ca CRON schedule = (CA duration - CA expiryWindow)`
2. The cert rotation cron schedule should be:  `Cert rotation CRON =min((Node duration-Node expiryWindow), (client duration-client expiryWindow))`
3. The CA expiry window should be greater than the Cert rotation cron schedule:  `CA expiryWindow > Cert rotation CRON schedule`

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+ 
+* 3 empty secret will be created in Helm chart for `cockroachdb-ca`, `cockroachdb-node` and `cockroachdb-root` if `tls.certs.generate.enabled`
+ is set.
+  * Data to these secrets will be populated in the `pre-install` job.
+  * In case CA is provided by the user, then `cockroachdb-ca` secret is skipped.
+  * Annotation is set on all the secrets created by CockroachDB; eg: `managed-by: cockroachdb`
+ 
+* A cronjob will be created in Helm chart when `tls.certs.generate.rotateCerts` is set.
+  * This cronjob will run periodically to rotate the certificates.
+  * The schedule of the cronjob will be of minimum `clientCertDuration` and `nodeCertDuration` minus some delta.
```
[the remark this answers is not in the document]
prafull01: To make cronjob period configurable we will add an entry in helm values.yaml as `expiryWindow` for each type of certificate.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
keith-mcclellan: there are two patterns here, rotating node certs that share a CA, and rotating a CA.  The first is a rolling restart, the latter is a full (parallel) restart

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
keith-mcclellan: My only comment here is that we need a way to handle arbitrary user/client certs - we always generate a root cert, but if we are using tls auth each user needs their own cert.  Ideally, we could run a job to trigger the generation of an arbitrary user cert (e.g. `karl`) and have it also covered by this cron job.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
keith-mcclellan: I would rather implement a rolling restart than a SIGHUP because managing sidecar containers would be a large increase in the scope of work here.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
DuskEagle: +1, we should have a different SA for generating these certificates than for running CRDB nodes. It's good for the service account which runs CRDB nodes to have as few privileges as needed for it to run.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
DuskEagle: I don't think a sidecar container is actually that much more work? We have the code for this in Cockroach Cloud, and we could easily open-source the few lines of code that it takes to do this. A SIGHUP is less disruptive to the cluster than restarting every node.

If either solution was the same amount of work, would you agree that a SIGHUP would be better?

keith-mcclellan: the problem is that we then have to distribute and maintain this additional sidecar container, including releasing new versions of it if there's a security patch, etc. It's the O&M overhead that I'm concerned about more than the actual work adding it to the helm chart itself. Does that make sense @DuskEagle ?

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+ 
+* Check if CA secret is empty
+  * if yes, [Generate-CA](#generate-ca), [Generate-Node-Cert](#generate-node-cert) and [Generate-Client-Cert](#generate-client-cert); return
+  * if not, [Validate Secret Annotations for CA](#validate-secret-annotations)
+    * if valid, CA is intact
+    * if not, [Annotate-CA](#annotate-ca), [Generate-Node-Cert](#generate-node-cert) and [Generate-Client-Cert](#generate-client-cert); return
+* Check if [CA requires rotation](#check-cert-for-regeneration)
+   * if yes, follow [Rotate-CA](#rotate-ca); return
+* Check if Node or Client certificates [needs to be regenerated](#check-cert-for-regeneration):
+   * if yes, follow [Generate-Node-Cert](#generate-node-cert) and [Generate-Client-Cert](#generate-client-cert)
```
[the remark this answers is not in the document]
prafull01: We can add the retries on the job failure. Also, Instead of spatting the failures in logs only, we can add the custom events on the cronJob about each operation we are performing. This would give user a better understanding of the cronjob when user runs `kubectl describe cronjob`

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
prafull01: Updated the RFC doc with the latest changes as per our discussions. You can see the updated changes in this new [commit](https://github.com/cockroachdb/helm-charts/pull/80/commits/1f8b4555bc1843dbce07d7f3a8afd186b3c13336)

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+ 
+CockroachDB user needs a default mechanism of cert management which should work on all the k8s distributions without the need for manual intervention. While cert-manager fits into the requirement, it makes CockroachDB deployment dependent on external application. This new method of cert-management satisfies the user requirement without the need for 3rd party software like cert-manager.
+ 
+## Goals
+ 
+* Helm install command should be self-sufficient to launch the CockroachDB cluster in a secure mode.
+ 
+* Dependency on external cert-manager should not be mandatory for creating CockroachDB cluster in a secure mode.
+ 
+* Manual steps should not be required for creating the CockroachDB cluster in a secure mode.
```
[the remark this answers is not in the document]
prafull01: Updated the goal in the document to deprecate the k8s CA signer

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
madhurnawandar: Updated RFC doc to not to remove SA, Role and RoleBinding

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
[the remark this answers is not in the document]
madhurnawandar: Cronjob schedule will not be taken from user but will be based on a formula using cert Duration and expiryWindow

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+  * Annotation will not be found, so this is a new CA scenario.
+* User had not given CA previously, but now has given the CA:
+  * Annotation will not be found, so this is a new CA scenario
+* If the user changes the duration of CA:
+  * Identify and compare using existing annotation on CA secret and current value, this will be a case for certificate rotation. This will only be the case when the CA is managed by CockroachDB.
+* User changes the duration of all certificates:
+  * Compare old and new CA duration from CA secret annotation values and current value. This will be a case for certificate rotation.
+  * Rotate CA certificate and add an annotation on CA secret with the date of rotation.
+  * Add annotations on Node and Client certs specifying the new expected duration and `to-be-rotated: true`.
+  * These secret certificates will be renewed in the next cron cycle and `to-be-rotated:true` and expected duration annotations will be removed.
```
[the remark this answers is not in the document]
madhurnawandar: This annotation will be generated by our code and is no user input. We have a separate helm input that decides if certs are to be rotated or not. Cron jobs will only be created in case cert-rotation is set to True via helm input

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
```diff
@@ -0,0 +1,243 @@
...
+ 
+#### Generate-Node-Cert
+* A Node certificate will be generated signed by the generated CA or user custom CA.
+* Expiry of the certificate will be driven by the Helm value for `tls.certs.generate.nodeCertDuration`, passed as env variable
+* Contents of the Node certificate will be stored in the Node secret `cockroachdb-node`.
+* An annotation `managed-by: cockroachdb` will be added on secret
+* Node secret will be patched with annotation `resourceVersion` with the value of its current resourceVersion
+* Node secret will be patched with annotation `creationTime` and `duration` with current UTC time and value from `tls.certs.generate.nodeCertDuration`
+ 
+#### Generate-Client-Cert
```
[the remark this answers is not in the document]
abhisek-dwivedi: We will create a helm plugin that can be used to generate any user's certificate apart from the root user certificate. Once a new user's certificate is created, it will be rotated/handled by the existing cert-rotation cronjobs.

docs/RFCs/19042021_Cert_Management_In_CRDB_Helm_Chart.md
DuskEagle: I see your reasoning. I still think we should be providing a way to reload certificates when they change on disk, either via a sidecar container or by native functionality in CRDB. But I'm fine with tabling this discussion and leaving it as a separate item from this RFC.

Issue #59, opened 2021-03-17
Use correct ingress apiVersion and structure depending on Kubernetes version


JorritSalverda:
Starting with kubernetes 1.19+ ingress is finally out of beta and needs a new `apiVersion` but also a slightly different structure. It has to include a `pathType` property for each path and the `backend` structure has changed. See https://kubernetes.io/docs/concepts/services-networking/ingress/.

JorritSalverda:
@JuanLeon1 thanks for merging the PR.

Unfortunately it turns out my own changes break things. Trying to run this I get

```
Error: Failed to render chart: exit status 1: Error: template: cockroachdb/templates/ingress.yaml:43:57: executing "cockroachdb/templates/ingress.yaml" at <.Capabilities.KubeVersion.GitVersion>: can't evaluate field Capabilities in type interface {}
```

Probably because this is nested with in a loop. I'm going to have a fix ready asap.

JorritSalverda:
https://github.com/cockroachdb/helm-charts/pull/74 should fix this issue. Sorry about this.

PR #81, opened 2021-04-20
fix ingress yaml invalid structure due to missing newlines

JorritSalverda:
With version 5.1.4 `helm template --set ingress.enabled=true cockroachdb` fails due to newlines getting removed too eagerly. This fixes the issue.

A test that would catch the regression would be useful. By running something like

```
helm template \
  --set conf.store.enabled=true \
  --set ingress.enabled=true \
  --set serviceMonitor.enabled=true \
  --set tls.enabled=true \
  --set networkPolicy.enabled=true \
  --set iap.enabled=true \
  --set iap.clientId=abc \
  --set iap.clientSecret=abc \
  cockroachdb
```

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 5.1.4
+version: 5.1.5
 appVersion: 20.2.7
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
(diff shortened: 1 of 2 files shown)


JorritSalverda:
@keith-mcclellan @DuskEagle I kept expecting CI/CD to catch the errors I introduced, my apologies for this. This time around i did test and verify with the above `helm template` command that the ingress template renders correctly.

I contributed to the grafana chart before, they use https://github.com/helm/chart-testing for testing the chart more thoroughly for PRs. Perhaps useful to have a look at.

JorritSalverda:
Fixes regression caused by https://github.com/cockroachdb/helm-charts/issues/59

Issue #82, opened 2021-04-20
No matches for kind "Ingress" in version "networking.k8s.io/v1" prior to Kubernetes 1.19


JorritSalverda:
Unfortunately the check `$.Capabilities.APIVersions.Has "networking.k8s.io/v1"` is successful for Kubernetes 1.17, but then fails applying the generated manifests with error

```
no matches for kind "Ingress" in version "networking.k8s.io/v1"
```

Issue https://github.com/kubernetes/kubernetes/issues/90077 show that the `Ingress` object isn't indeed available even though the new `apiVersion` is.

Looking at how the Grafana chart does this - see https://github.com/grafana/helm-charts/blob/main/charts/grafana/templates/ingress.yaml#L7 - it turns out to need `$.Capabilities.APIVersions.Has "networking.k8s.io/v1/Ingress"` to accomplish what I was trying to do.

I'll issue yet another PR and trying to figure out how I'll test this prior to it getting merged :S

PR #83, opened 2021-04-20
include Ingress object in APIVersions check

JorritSalverda:
Fixes https://github.com/cockroachdb/helm-charts/issues/82.

I've verified this structure to work in https://github.com/estafette/estafette-ci-api/blob/main/helm/estafette-ci-api/templates/ingress.yaml#L2-L8 by running it against both kubernetes 1.17 and 1.19 in my build pipeline successfully.

It's actually elasticsearch' Helm charts that do more extensive testing (not grafana as I mentioned before), see https://github.com/elastic/helm-charts/blob/master/CONTRIBUTING.md#testing.

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 5.1.5
+version: 5.1.6
 appVersion: 20.2.7
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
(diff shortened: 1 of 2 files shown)


JorritSalverda:
@keith-mcclellan 3rd time's a charm? :|

keith-mcclellan (approved):
LGTM

keith-mcclellan:
fwiw we are discussing adding some additional testing @JorritSalverda so thanks for pointing that out.  We lost some of our test harness when we moved from the old helm chart repo to this one.

JorritSalverda:
Thanks, for merging. And now... drumroll.. it finally works. Phew, that took a couple of attempts too many, hope it didn't negatively impact all the chart users out there while it was broken.

PR #84, opened 2021-04-22
Adding terratest and an initial test on a simple template

udnay:
Draft to see how terratest could be structured. I'm not sure if this structure makes sense or not but wanted to have something to discuss.


test/cockroachdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/stretchr/testify/require"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+func Test_Role(t *testing.T) {
```
keith-mcclellan, line 13: godocs / comments please

test/cockroachdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+	helmChartPath, err := filepath.Abs("../cockroachdb")
+	require.NoError(t, err)
+	require.NotNil(t,helmChartPath)
+
+	t.Run("Test TLS Enabled but certs not provided", func(t *testing.T) {
```
keith-mcclellan, line 18: We need to validate this in combination with the k8s version - this will work on older versions of k8s with the v1beta1 api but will fail on newer versions where the legacy unknown signer isn't available.  So this test needs to be more specific

test/cockroachdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+		require.Len(t, verbs, 1)
+		require.Contains(t, verbs, "get")
+	})
+
+	t.Run("Test TLS disabled", func(t *testing.T) {
+		options := &helm.Options{
+			SetValues: map[string]string{
+				"tls.enabled": "False",
+				"tls.certs.provided": "True",
```
keith-mcclellan, line 71: if you're testing TLS disabled, this line should also be false, and the db should be running in insecure mode

test/cockroachdb_test.go
keith-mcclellan, line 13: The below tests should be broken out into individual tests please, rather than rolled all together.

keith-mcclellan (changes requested):
In general, we need more godocs / comments.  Also, I generally prefer smaller, more atomic tests that are easier to troubleshoot.  This reads like a single test covering a bunch of different security modes.  Please correct me if I'm wrong about that.

DuskEagle (commented):
Can you add a README.md on how to run these tests? I would've thought it would be something like `go test ./test/...`, but the tests all fail for me with `error while running command: exit status 1; Error: could not find template templates/role.yaml in chart`. Are you running something different?

test/cockroachdb_test.go
DuskEagle: The DB will run in insecure mode when tls.enabled is false?

https://github.com/cockroachdb/helm-charts/blob/f37ff112fbbee969d64dc19526226c62bb9651fc/cockroachdb/templates/statefulset.yaml#L224

This is something that a test on the statefulset template could ensure.

test/cockroachdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+	require.NotNil(t,helmChartPath)
+
+	t.Run("Test TLS Enabled but certs not provided", func(t *testing.T) {
+		options := &helm.Options{
+			SetValues: map[string]string{
+				"tls.enabled": "True",
+				"tls.certs.provided": "False",
+				"tls.certs.certManager": "False",
+			},
+			KubectlOptions: k8s.NewKubectlOptions("", "", "test-deployment"),
```
DuskEagle, line 25: Since this test is only rendering templates, not running them, I don't think `KubectlOptions` is needed, even though the terratest example includes it. The tests seem to pass without it.

udnay:
That’s how I run it. The last test does error out but the rest should pass.
I’ll check if I’m doing anything special

test/cockroachdb_test.go
DuskEagle: I think `t.Run` is fine, since this test is focused on rendering and validating the `role.yaml` template. I think we could also potentially use [table-driven tests](https://github.com/golang/go/wiki/TableDrivenTests) to make this even more idiomatic. WDYT?

test/cockroachdb_test.go
```diff
@@ -0,0 +1,82 @@
...
+		output := helm.RenderTemplate(t, options, helmChartPath, releaseName, []string{"templates/role.yaml"})
+		require.NotNil(t, output)
+
+		var role rbacv1.Role
+		helm.UnmarshalK8SYaml(t, output, &role)
+
+		verbs := role.Rules[0].Verbs
+
+		require.Len(t, verbs, 2)
+		require.Contains(t, verbs, "get")
```
DuskEagle, line 39: If you add a test like the below, it could help with making it more clear what resource the verbs are actually referring to in this test.

```
resources := role.Rules[0].Resources
require.Len(t, resources, 1)
require.Contains(t, resources, "secrets")
```

DuskEagle (commented):
I like the direction this is going!

It would be great if we could also get integration tests working at some point, which verified that the cockroachdb nodes became healthy. But just being able to validate components of the rendered templates is a big step forward.

build/tester/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM quay.io/helmpack/chart-testing:v3.3.1
+
+COPY --from=golang:1.15-alpine /usr/local/go/ /usr/local/go/
+ENV PATH="/usr/local/go/bin:${PATH}"
+
+# Install gcc.
+RUN apk add build-base
+
```
udnay, line 8: @jlinder Can you take a look at this and let me know what you think? Also is there a way to cache the go dependencies listed in `go.mod` without running `go test`? I'm thinking of something like `npm install`

DuskEagle: I think what you're looking for is `go mod vendor`. It will create a `vendor` directory which you can check-in to source control.

udnay:
I don’t want the dependencies to be in source control, I just think I can
cache the intermediate build where dependencies are pulled down since the
dependencies will change less frequently than the tests or charts
themselves. So far the only way I've managed to get the dependencies is
when I run `go test ./test`. I did try `go mod download` but because the
helm chart isn't actually a go module it doesn't seem to help.

DuskEagle:
Hmm, I'm not 100% sure what you're looking for. `go mod download` from the root of this repo downloads the `terratest` and `testify` dependencies to my `~/go/pkg/mod/` directory.

PR #86, opened 2021-04-28
Add existing imagePullSecrets support in addition to credentials

arzhr:
Add support of existing imagePullSecrets in addition to registry credential

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -33,13 +33,19 @@ spec:
     spec:
       restartPolicy: OnFailure
       terminationGracePeriodSeconds: 0
-    {{- if or .Values.image.credentials (and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) }}
+    {{- if or .Values.image.credentials .Values.image.imagePullSecrets (and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) (and .Values.tls.enabled .Values.tls.init.image.imagePullSecrets (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) }}
       imagePullSecrets:
       {{- if .Values.image.credentials }}
         - name: {{ template "cockroachdb.fullname" . }}.db.registry
       {{- end }}
+      {{- if .Values.image.imagePullSecrets }}
+        - name: {{ .Values.image.imagePullSecrets }}
+      {{- end }}
       {{- if and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager) }}
         - name: {{ template "cockroachdb.fullname" . }}.init-certs.registry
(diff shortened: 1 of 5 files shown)


DuskEagle (commented):
Can you add to [`README.md#Configuration`](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/README.md#configuration) what `imagePullSecrets` is?

cockroachdb/README.md
```diff
@@ -310,6 +310,7 @@ For details see the [`values.yaml`](values.yaml) file.
 | `image.repository`                                        | Container image name                                            | `cockroachdb/cockroach`                          |
 | `image.tag`                                               | Container image tag                                             | `v20.2.8`                                        |
 | `image.pullPolicy`                                        | Container pull policy                                           | `IfNotPresent`                                   |
+| `image.imagePullSecrets`                                  | Specify existing Image pull secret                              | `""`                                             |
```
DuskEagle, line 313: It's not clear to me how to use this from this description. Do we specify a K8s secret name? 

If so, it may help to include a hyperlink to some Kubernetes documentation on this feature, so others can know what that custom secret needs to contain.

cockroachdb/values.yaml
```diff
@@ -390,6 +391,7 @@ tls:
       repository: cockroachdb/cockroach-k8s-request-cert
       tag: "0.4"
       pullPolicy: IfNotPresent
+      imagePullSecrets: ""
```
DuskEagle, line 394: Document this setting too.

ckoehn:
@arzhr Is this something you will eventually work on again? We need that feature too and would be happy to take over/continue.

cockroachdb/templates/job.init.yaml
```diff
@@ -33,14 +33,20 @@ spec:
     spec:
       restartPolicy: OnFailure
       terminationGracePeriodSeconds: 0
-    {{- if or .Values.image.credentials (and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) }}
+    {{- if or .Values.image.credentials .Values.image.imagePullSecrets (and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) (and .Values.tls.enabled .Values.tls.init.image.imagePullSecrets (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) }}
       imagePullSecrets:
       {{- if .Values.image.credentials }}
         - name: {{ template "cockroachdb.fullname" . }}.db.registry
       {{- end }}
+      {{- if .Values.image.imagePullSecrets }}
```
mdomke, line 41: I think it might be confusing that the value is called `imagePullSecrets` (in plural form), but you can actually just specify a single secret. I would either do something like this 
```
    {{- if .Values.image.imagePullSecrets }}
      imagePullSecrets:
      {{- range .Values.image.imagePullSecrets }}
        - name: {{ . }}
      {{- end}}
    {{- end }}
```
or rename the variable to `imagePullSecret`. (I actually prefer the first solution).

udnay:
Closing as stale

Issue #87, opened 2021-04-30
Run init job for single node to create root user certificate


JorritSalverda:
In order to run a single node 'cluster' in secure mode it's useful to have the init job run so it can issue the CSR for the root client certificate. I've verified that the job runs fine for single node. It does log the following error but completes succesfully:

```
ERROR: cluster has already been initialized Failed running "init"
```

PR #88, opened 2021-04-30
Run init job for single node

JorritSalverda:
Fixes https://github.com/cockroachdb/helm-charts/issues/87

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 5.1.7
+version: 5.1.8
 appVersion: 20.2.8
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
 icon: https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png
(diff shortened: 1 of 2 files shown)


JorritSalverda:
@keith-mcclellan any chance this can get merged? Thx

keith-mcclellan:
there's a special start command for single-node cockroachdb... `cockroach start-single-node` that doesn't require init.  see https://www.cockroachlabs.com/docs/v20.2/cockroach-start-single-node.html.

keith-mcclellan (changes requested):
there's a special start command for single-node cockroachdb... cockroach start-single-node that doesn't require init. see https://www.cockroachlabs.com/docs/v20.2/cockroach-start-single-node.html.  So I'm not sure what the purpose of this PR is?

JorritSalverda:
It ensures the root client certificate gets generated so you can still use the single node 'cluster' in secure mode from a client application. Any other way to generate those client certificates easily from the Helm chart would work as well, but this is an easy and harmless way of doing so.

JorritSalverda:
@keith-mcclellan I've created an `init-cert` job instead that runs when tls is enabled and `single-node` is used. It feels like there's more to maintain though, so not sure what you prefer in order to generate a client certificate for the single-node option.

PR #91, opened 2021-05-22
sort storage params for stable helm upgrade

EladDolev:
subsequent `helm upgrade` may fail because the StatefulSet allows only handful of changes, and "cockroachdb.conf.store" function may return different output every time


EladDolev:
any chance somebody can review this ?
this thing causes random pod restart on every helm apply...

sergeyshaykhullin:
Seems valid, but i use store config half a year and upgrade never failed

EladDolev:
Upgrade should not fail, but may cause unnecessary pods restart

PR #107, opened 2021-06-04
Add custom liveness and readiness probes

sergeyshaykhullin:
My crdb nodes sometimes failing with readiness probes or timeouts

I added custom probes to:
- Decrease probe interval, 5s seems very often
- Increase failure threshold 2 failures very low
- Increase timeout seconds - 1 second by default


cockroachdb/values.yaml
```diff
@@ -228,6 +228,9 @@ statefulset:
     #   cpu: 100m
     #   memory: 512Mi
 
+  customLivenessProbe: {}
+  customReadinessProbe: {}
```
DuskEagle, line 232: Can you add these values to [cockroachdb/README.md#Configuration](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/README.md#configuration), and provide an example here in values.yaml of what a customLivenessProbe might look like?

DuskEagle:
Looks good to me.

PR #109, opened 2021-06-11
Add the CI workflows for cockroachdb

prafull01:
Added jobs for following workflows:
- go lint
- helm lint
- unit test
- build self-signer utility
- integration test

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,104 @@
+name: Helm Chart Package CI
+on:
+  push:
+    branches:
+      - 'master'
+      - 'cert-manager-feature-branch'
+    tags:
+      - 'v*.*.*'
+  pull_request:
+    branches:
+      - 'master'
+      - 'cert-manager-feature-branch'
+
+jobs:
+  # pre job run golangci-lint
+  go-lint:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,105 @@
...
+    steps:
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to GCR
+        uses: docker/login-action@v1
+        with:
+          registry: gcr.io
+          username: _json_key
+          password: ${{ secrets.GCR_HELM_CHART_SA_JSON` }}
```
udnay, line 76: Need to get rid of the ` character.

Are you sure you can use the docker login flow and just pass in the JSON cert file? I don't think this will work, have you tested it?

udnay (commented):
Can you please verify that this works as expected?

.github/workflows/ci.yaml
prafull01: I have fixed this typo and verified that it works on GCR repo. Link for my fork where it is working: https://github.com/prafull01/helm-charts/pull/4

udnay (approved):

PR #110, opened 2021-06-17
Add template tests of self signed certificates for the helm charts

prafull01:
Add the helm template tests for the self signer utility.
Fix the validations and values-schema.json identified using those tests.
Add table driven tests to optimize the test code


tests/template/cockroachdb_helm_template_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/logger"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+)
+
+// TestHelmSelfCertSignerTemplateRenderedSuccessfully contains the successful rendered objects by the cockroachdb helm
+// chart with selfSigner enabled.
+func TestHelmSelfCertSignerTemplateRenderedSuccessfully(t *testing.T) {
```
udnay, line 22: Can we have some scenarios where we check more than the namespace and check that parts of the template guarded by `if/then` statements are what we expect?

prafull01: I have added the if else check in the `statefulset.yaml` and `job.init.yaml`. These two files changes based on selfSigner.enabled value. 
Also, added tests for disable the selfSigner and check the self signer resources are not templated.
Implemented table driven tests to optimize the test code.

udnay (approved):
LGTM

Issue #103, opened 2021-05-31
Can log configuration be added to the chart?


Temikus:
Currently the chart doesn't allow operating the `--log <string>` flag which is important to us as we've been waiting for a long time for JSON logging and want to get rid of ngrok parsers ASAP.

Thank you.

sergeyshaykhullin:
I think we can:
- [ ] Add `conf.logging` section with `enabled: false/true`
- [ ] Add configmap from logging section
- [ ] Mount it to pods
- [ ] Pass file to `--log-config-file`

More details: https://www.cockroachlabs.com/docs/v21.1/configure-logs

sergeyshaykhullin:
Oh, no, you can't... Because `--logtostderr` is always applied...

PR #113, opened 2021-06-17
New logging configuration

sergeyshaykhullin:
# Problem
This needs to specify `--log` or `--log-config-file`. New logging api can't be used with old one

Details: https://github.com/cockroachdb/helm-charts/issues/103

# Changes
Do not append `--logtostderr` and add secret + mount + --log-config-path args if `.Values.log.enabled`


Temikus:
Hey folks, this is blocking us a bit as ngrok parsing keeps breaking on corner cases - any way I can help move this along?

123BLiN:
In our case we need slow log with query time correctly parsed

sergeyshaykhullin:
My next steps is to add cm for log config, mount it to all pods and pass log file path flag if enabled

It would be better and more... Native and transparent then operating on strings and do helm conversions

Also you will not have a very long logs config args on process start

But the main bottleneck is a review... I can't speed it up ;c

udnay:
@sergeyshaykhullin can you add some test cases for this change please? You can see an example here https://github.com/cockroachdb/helm-charts/blob/master/tests/template/cockroachdb_helm_template_test.go

sergeyshaykhullin:
@udnay Seems like done

udnay (approved):
LGTM

PR #121, opened 2021-07-04
Converting the hours into valid cron schedule as per kubernetes

prafull01:
Kubernetes doesn't support cron schedule like `0 */624 * * *` where the hours are greater than 24, days are greater than 31 and months are greater than 12. Hence added a mechanism to convert cron schedule from hours to supported cron schedule in hours, days, months and years.

### DIFF
--- a/cockroachdb/templates/_helpers.tpl
+++ b/cockroachdb/templates/_helpers.tpl
@@ -96,15 +96,71 @@ Define the default values for the certificate selfSigner inputs
 {{- end -}}
 
 {{/*
-Define the cron schedules for certificate rotate jobs
+Define the cron schedules for certificate rotate jobs and converting from hours to valid cron string.
+We assume that each month has 31 days, hence the cron job may run few days earlier in a year.
 */}}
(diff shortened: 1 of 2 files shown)


cockroachdb/templates/_helpers.tpl
```diff
@@ -96,15 +96,48 @@ Define the default values for the certificate selfSigner inputs
 {{- end -}}
 
 {{/*
-Define the cron schedules for certificate rotate jobs
+Define the cron schedules for certificate rotate jobs and converting from hours to valid cron string
```
udnay, line 99: Add a comment here that we assume each month has 31 days and as a result the cron job may fire earlier than specified.

31*12 = 371 day, which is probably fine

prafull01: I have updated the comment. Also, I have improved the logic which creates a cron schedule like `0 0 */28 * *` which means the cron will run every 28 days.

Also, added the helm template tests for the verification

udnay (approved):
LGTM

Issue #99, opened 2021-05-25
certGenerator: Client/Node Cert renewal module


madhurnawandar:
Add a wrapper on top of cert generation module for client and node certificates,
to renew these certificates in case CA is renewed

PR #122, opened 2021-07-07
Added support for CA, node and client cert rotation #98 #99

### LINKED ISSUE #98: certGenerator: CA renewal module
Add module to renew CA and generate SIGHUP to add nodes for renewal
Usable in the case of self-managed CA.

abhishekdwivedi3060:
closes #98 #99 

* Added support for CA, node and client cert rotation
* Added corresponding test-cases

### DIFF
--- a/cmd/self-signer/rotate.go
+++ b/cmd/self-signer/rotate.go
@@ -17,9 +17,10 @@ limitations under the License.
 package self_signer
 
 import (
-	"fmt"
-
 	"github.com/spf13/cobra"
+	"log"
+	"os"
+	"time"
 )
 
 // rotateCmd represents the rotate command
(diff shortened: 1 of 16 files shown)


cmd/self-signer/rotate.go
```diff
@@ -30,10 +31,62 @@ var rotateCmd = &cobra.Command{
...
+	rotateCmd.Flags().StringVar(&caCron, "ca-cron", "", "cron of the CA certificate rotation cron")
+	rotateCmd.Flags().StringVar(&nodeAndClientCron, "node-client-cron", "", "cron of the node and client certificate rotation cron")
+
+	rotateCmd.Flags().StringVar(&readinessWait, "readiness-wait", "30s", "readiness wait for each replica of crdb cluster")
+
 }
 
 func rotate(cmd *cobra.Command, args []string) {
-	fmt.Println("rotate called")
+	if clientFlag && nodeFlag && caFlag {
```
udnay, line 55: Doesn't this need to be 
`(clientFlag || nodeFlag) && caFlag` otherwise client and CA could be rotated together.

cmd/self-signer/rotate.go
```diff
@@ -30,10 +31,62 @@ var rotateCmd = &cobra.Command{
...
+		log.Panic("CA, Node and client can't be rotated at the same time. Only CA or (Node and Client) can be " +
+			"rotated at a time")
+	}
+
+	if !clientFlag && !nodeFlag && !caFlag {
```
udnay, line 60: /nit
`!(clientFlag || nodeFlag || caFlag)`

pkg/generator/generate_cert.go
```diff
@@ -117,6 +123,12 @@ func (rc *GenerateCert) Do(ctx context.Context, namespace string) error {
 		return errors.Wrap(err, msg)
 	}
 
+	// In the case of rotate CA, skip node and client certificate rotation
+	if rc.RotateCACert {
+
```
udnay, line 128: delete line

pkg/generator/generate_cert.go
```diff
@@ -167,146 +179,194 @@ func (rc *GenerateCert) generateCA(ctx context.Context, CASecretName string, nam
...
+				return nil
+			}
+		}
+
+		logrus.Infof("Node secret [%s] is found in ready state, skipping Node cert generation", nodeSecretName)
+		return nil
 	}
 
-	logrus.Infof("Generated and saved node key and certificate in secret [%s]", nodeSecretName)
+	return generate()
```
udnay, line 368: Why are we doing the closure here? And also above? Can't we just inline the function?

udnay (commented):
A couple of comments but looks good

cmd/self-signer/rotate.go
abhishekdwivedi3060: Other cert rotation will be skipped if `caFlag` is set, so client and CA won't be rotated together. But you are right we should handle above case and error out instead.
Fixed

cmd/self-signer/rotate.go
abhishekdwivedi3060: Done

pkg/generator/generate_cert.go
abhishekdwivedi3060: Done

pkg/generator/generate_cert.go
abhishekdwivedi3060: We are not manipulating any values in the function so kept it as closure. Changed it to inline function for better readability.

udnay (approved):

PR #126, opened 2021-07-12
Converting the hours into valid cron schedule as per kubernetes

prafull01:
The kubernetes cron doesn't allow to run a cron job for more than a year. Hence added a fix to run in combination of months.

For example: A CA cron with 5 years expiry and 2 months expiry window, means a total of 58 month when the cron should be scheduled. The cron will schedule every 11th month making the rotation happen at  55th month.


prafull01:
To Review: @udnay

prafull01:
Closing this, as this fix I have integrated with my e2e tests PR.

Issue #100, opened 2021-05-25
certGenerator: e2e testing


madhurnawandar:
Scenario identification and implementation

PR #127, opened 2021-07-13
Add E2e tests for the cockroachdb helm install

prafull01:
- Add e2e tests for the cockroachdb helm install.
- The kubernetes cron doesn't allow to run a cron job for more than a year. Hence added a fix to run in combination of months.

For example: A CA cron with 5 years expiry and 2 months expiry window, means a total of 58 month when the cron should be scheduled. The cron will schedule every 11th month making the rotation happen at 55th month.


tests/testutil/require.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	return count
+}
+
+func RequireCertificatesToBeValid(t *testing.T, crdbCluster CockroachCluster) {
+	t.Log("Verifying the Certificates")
+	kubectlOptions := k8s.NewKubectlOptions("", "", crdbCluster.Namespace)
+
+	// Get the CA certificate secret and load the ca cert
+	caSecret := k8s.GetSecret(t, kubectlOptions, crdbCluster.CaSecret)
```
udnay, line 199: Maybe put these inside of the `if` below on line 202

prafull01: We need to get the CA cert in both the cases, if ca is user provided or not. As we are using this CA certificate to verify the node certificates are signed by this CA or not at line 217.

cockroachdb/templates/_helpers.tpl
```diff
@@ -97,70 +97,70 @@ Define the default values for the certificate selfSigner inputs
 
 {{/*
 Define the cron schedules for certificate rotate jobs and converting from hours to valid cron string.
-We assume that each month has 31 days, hence the cron job may run few days earlier in a year.
+We assume that each month has 31 days, hence the cron job may run few days earlier in a year. In a cron schedule,
```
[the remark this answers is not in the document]
prafull01: The problem with running a cron everyday means over a 5 year expiry CA certificate we would run cron job every day and the certificate rotation will happen just a day before its expiry. The CA certificate rotation should happen months before actual expiry. Even putting the every month cron run means running a cron over 58 times before actual rotation happens. 

With this fix, the cron run will happen only 4 times before the actual rotation of CA certification happens.

keith-mcclellan (approved):
LGTM

prafull01:
Issue #100 is fixed by this PR.

PR #130, opened 2021-07-15
Issue 129: Added post-delete helm job to delete dangling secret

abhishekdwivedi3060:
* Added post-delete helm job to delete dangling secret
* Added unit test-case

### DIFF
--- a/cmd/self-signer/cleanup.go
+++ b/cmd/self-signer/cleanup.go
@@ -32,7 +32,7 @@ var cleanupCmd = &cobra.Command{
 	Use:   "cleanup",
 	Short: "cleanup cleans up the secrets generated using self-signer utility",
 	Long:  `cleanup sub-command cleans up the secrets i.e. node, client and CA secrets generated using self-signer utility`,
-	Run:   cleanup,
+	RunE:  cleanup,
 }
 
 var namespace string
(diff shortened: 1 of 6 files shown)


cmd/self-signer/cleanup.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func cleanup(cmd *cobra.Command, args []string) {
+
+	stsName, exists := os.LookupEnv("STATEFULSET_NAME")
+	if !exists {
+		log.Fatal("Required STATEFULSET_NAME env not found")
+	}
+
+	resource.Clean(ctx, cl, namespace, stsName)
```
udnay, line 55: Where does the `ctx` come from?

cmd/self-signer/cleanup.go
```diff
@@ -0,0 +1,57 @@
...
+	logrus.Info("Successfully cleaned up dangling resources")
```
udnay, line 56: I think `resource.Clean` should return an `error` or `bool` indicating if it cleaned all the secrets without an error. Otherwise the log message is confusing.

udnay (commented):
Couple of questions but looking good

cmd/self-signer/cleanup.go
abhisek-dwivedi: This is a global context that is initialized in the package init function and is passed to all sub-commands

cmd/self-signer/cleanup.go
abhisek-dwivedi: Done

cmd/self-signer/cleanup.go
abhishekdwivedi3060: This is a global context that is initialized in the package init function and is passed to all sub-commands

cmd/self-signer/cleanup.go
abhishekdwivedi3060: Done

udnay: I dont think we need to use `RunE` here, especially if we're just always returning `nil`

abhishekdwivedi3060: Done

PR #134, opened 2021-07-19
Proposal: Optionally expose node locality (region, zone, host) to pod

scottcrossen:
Discussion: Has this been investigated before? I imagine that most people want these locality tags in multi-region setups. This PR adds them from [well-known node labels](https://kubernetes.io/docs/reference/labels-annotations-taints/). I'm happy to transfer ownership of my proof-of-concept `initContainer` that [mounts](https://github.com/scottcrossen/kube-node-labels) the node labels in the statefulset.

TLDR: This pullrequest automatically adds the CLI start flag values
```
--locality=region=us-west-2,zone=us-west-2a,host=myhost
```


scottcrossen:
Rebased

scottcrossen:
Rebased

cockroachdb/templates/job.init.yaml
```diff
@@ -16,9 +16,9 @@ metadata:
   {{- with .Values.labels }}
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  annotations:
```
sergeyshaykhullin, line 19: I don't like this approach, helm hooks annotations should always be present or someone who wants to add new annotations will override helms one

sergeyshaykhullin: You can add something like extraJobAnnotations and list after helm hooks

scottcrossen: Hi! The purpose of this was to make it possible to remove the annotations.

I've moved this discussion into a new PR: https://github.com/cockroachdb/helm-charts/pull/206

scottcrossen:
Rebased PR

cockroachdb/templates/job.init.yaml
sergeyshaykhullin: Got it. Thanks!

PR #137, opened 2021-07-20
Add client/node and CA certificate rotation tests 

prafull01:


tests/e2e/rotate/cockroachdb_rotate_certs_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Deploy the chart using `helm install`. Note that we use the version without `E`, since we want to assert the
+	// install succeeds without any errors.
+	helm.Install(t, options, helmChartPath, releaseName)
+
+	// Next we wait for the service endpoint
+	serviceName := fmt.Sprintf("%s-cockroachdb-public", releaseName)
+
+	k8s.WaitUntilServiceAvailable(t, kubectlOptions, serviceName, 30, 2*time.Second)
+
```
keith-mcclellan, line 84: Can we get some more comments breaking down what we're doing below please?

keith-mcclellan: I'd also prefer more verbose logging for this test in case it fails and we need to troubleshoot

keith-mcclellan (changes requested):
overall looks good, but I'd like some more comments and some more verbose logging in cockroachdb_rotate_certs_test.go

tests/e2e/rotate/cockroachdb_rotate_certs_test.go
prafull01: Resolved

keith-mcclellan (approved):
LGTM thanks!

PR #139, opened 2021-07-21
Run Copy Certs for Cert Manager

### LINKED ISSUE #78: copy-certs init container not created when using cert-manager
I would expect when using `cert-manager` that setting `tls.certs.certManager` is enough. However, the `copy-certs` init container is not created since `tls.certs.provided` has to be `true` as well. Is this intended? If yes, the documentation should explicitly state that.

nikkomiu:
This should close #78 by adding the init container when cert manager is being used.


udnay:
Can you rebase this please and then we'll take a look

PR #140, opened 2021-07-22
Add the support of custom user client cert generation

prafull01:
User can generate client certificates for any user using following job:
```
apiVersion: batch/v1
kind: Job
metadata:
  name: user-cert-generate
spec:
  backoffLimit: 1
  completions: 1
  parallelism: 1
  template:
    spec:
      restartPolicy: Never
      containers:
      - args:
        - generate
        - --client-only
        env:
        - name: USER_NAME
          value: prafull              ## Provide the user name to which you want to create client certificates
        - name: CA_SECRET
          value: crdb-cockroachdb-ca-secret  ## CA secret where the CA is stored for CRDB
        - name: NAMESPACE
          valueFrom:
            fieldRef:
              fieldPath: metadata.namespace
        image: gcr.io/cockroachlabs-helm-charts/cockroach-self-signer-cert:dc73bb7f9a40013696fa342af78872fdaa927163
        imagePullPolicy: Always
        name: user-cert-job
      serviceAccountName: crdb-cockroachdb-self-signer
```

### DIFF
--- a/pkg/generator/generate_cert.go
+++ b/pkg/generator/generate_cert.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+	"os"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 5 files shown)


pkg/generator/generate_cert.go
```diff
@@ -145,32 +146,47 @@ func (rc *GenerateCert) Do(ctx context.Context, namespace string) error {
 	return nil
 }
 
+func (rc *GenerateCert) ClientCertGenerate(ctx context.Context, namespace string) error {
```
keith-mcclellan, line 149: Can I get some comments here please

pkg/generator/generate_cert.go
```diff
@@ -517,3 +542,25 @@ func (rc *GenerateCert) UpdateNewCA(ctx context.Context, namespace string) error
 	}
 	return nil
 }
+
+func (rc *GenerateCert) LoadCASecret(ctx context.Context, namespace string) error {
```
keith-mcclellan, line 546: Please put a code comment here as well

keith-mcclellan (changes requested):
Please add some additional comments for the new functions, otherwise LGTM

pkg/generator/generate_cert.go
prafull01: done

pkg/generator/generate_cert.go
prafull01: done

keith-mcclellan (approved):

PR #141, opened 2021-07-22
Add TLS to Service Monitor Endpoint

nikkomiu:


mprimeaux:
Just curious as to why this was closed without merging?

PR #143, opened 2021-07-23
add flag to switch to cert-manager.io/v1 apiVersion

### LINKED ISSUE #142: Add flag to allow usage of cert-manager.io/v1 apiVersion
We are in the special situation that for a transition period we need to run K8s 1.15.11 with cert-manager 1.1 which requires the usage of cert-manager's legacy CRD set, which is using `cert-manager.io/v1` apiVersion, but lacks the features to accept older apiVersions and automatically convert them (only supported from K8s 1.16 upwards). 
For that matter, we'd need a flag added to the helm chart which allows to use `cert-manager.io/v1` Certificates already.

I will open a PR to propose that change as an optional switch to allow the usage of the newer API Version

christianhuening:
This adds a flag `.tls.certs.useCertManagerV1CRDs` defaulting to `false`. If set to `true` the cert-manager resources will be rendered using `cert-manager.io/v1` .

Fixes #142

### DIFF
--- a/cockroachdb/templates/certificate.client.yaml
+++ b/cockroachdb/templates/certificate.client.yaml
@@ -23,8 +23,14 @@ spec:
     - digital signature
     - key encipherment
     - client auth
+{{- if .Values.tls.certs.useCertManagerV1CRDs }}
+  privateKey:
+    algorithm: rsa
+    size: 2048
+{{- else }}
   keySize: 2048
   keyAlgorithm: rsa
+{{- end }}
   commonName: root
   organization:
     - Cockroach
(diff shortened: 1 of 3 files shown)


udnay (approved):
LGTM

daniel-crlabs:
Has this been released or do we know what release will contain this fix? Please let us know as we have a customer who is impacted and would like an update.

daniel-crlabs:
@thtruo or @udnay Do you know if this is now available? Can the customer use the latest helm version and will it have this feature?

thtruo:
For continuity: Daniel caught up with Yandu - this is not available yet; the team is about to merge a large refactor and will need to test before making this available

daniel-crlabs:
Thank you both.

PR #144, opened 2021-07-23
Fix the SHA value of E2e tests

prafull01:
Need to add a env variable TAG which will have PR commit sha.
Currently in tests we are using default GITHUB_SHA which have the master commits sha not the PR commits sha and hence the docker build happen with different sha and tests run with different sha


keith-mcclellan (approved):
LGTM

PR #145, opened 2021-07-29
Add a new approach to our CI

prafull01:
- Add a build job which will be triggered only on the successful PR merge or push on the master.
- Add a step to build the image as a part of E2E.


keith-mcclellan (approved):
LGTM

PR #146, opened 2021-07-30
Remove the support of kubernetes CA certificate signing

prafull01:


cockroachdb/templates/job.init.yaml
```diff
@@ -33,57 +34,21 @@ spec:
...
       {{- if .Values.image.credentials }}
         - name: {{ template "cockroachdb.fullname" . }}.db.registry
       {{- end }}
-      {{- if and .Values.tls.enabled .Values.tls.init.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager) }}
-        - name: {{ template "cockroachdb.fullname" . }}.init-certs.registry
+      {{- if and .Values.tls.enabled .Values.tls.selfSigner.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager) }}
+        - name: {{ template "cockroachdb.fullname" . }}.self-signed-certs.registry
       {{- end }}
     {{- end }}
-    {{- if and .Values.tls.enabled (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager) }}
```
keith-mcclellan, line 45: Isn't this code also being used for the cert-manager integration thats already in the chart?

keith-mcclellan (changes requested):
I'm concerned that these changes may break the cert-manager integration that the helm chart already has. Can we confirm that isn't the case please?

cockroachdb/templates/job.init.yaml
prafull01: This condition implies that if tls is enabled, however when certs are not provided and certManager is also disabled. It means you need to sign it using kubernetes CA. So this initContainer doesn't come into picture with certManager enabled

Nonetheless, I will verify that the certManager integration is working fine with this change and update it here

keith-mcclellan: sorry I misread thanks!

PR #154, opened 2021-08-10
Add certificate migration e2e test #131

### LINKED ISSUE #131: TLS Cert migration testing
At a high level, we need to verify the following workflows:

1 migrate customer from k8s CA signed certs to self signed certs
2 migrate customer from custom certs to self signed certs
3 migrate customer from self signed certs to custom certs
4 migrate customer from self signed certs to cert manager certs

Most of these can be validated manually, but I would pick either #2 or #3 and make a permanent test to make sure we don't break this functionality in the future.

abhishekdwivedi3060:
closes #131 
* Added migration test from "Bring Your Own Certificate" method to self-signer method

### DIFF
--- a/pkg/generator/generate_cert.go
+++ b/pkg/generator/generate_cert.go
@@ -390,7 +390,7 @@ func (rc *GenerateCert) generateClientCert(ctx context.Context, clientSecretName
 
 	user, userExist := os.LookupEnv("USER_NAME")
 	if !userExist {
-		user = "root"
+		user = security.RootUser
 	} else {
 		clientSecretName = fmt.Sprintf("%s-client-secret", user)
 	}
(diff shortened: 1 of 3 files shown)


udnay:
Looks like the failure might be legit
> UPGRADE FAILED: cannot patch "crdb-test-cockroachdb" with kind StatefulSet: StatefulSet.apps "crdb-test-cockroachdb" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden

abhishekdwivedi3060:
The failure was because of a typo. Fixed that.

abhishekdwivedi3060:
While upgrading, the default self-signer method is used and in the statefulset template condition, the node secret names are different for BYOC and self-signer based on the `.Values.tls.certs.selfSigner.enabled` . So IMO no need to delete the user provided secrets to test migration.

udnay (approved):
LGTM

PR #155, opened 2021-08-16
Improve the CI and build workflows of cert-manager utility

prafull01:
Changed the workflow of CI:
- Added a job which will check the tag is increased or not if the cert utility has been changed.
- Post job will build the new tag mentioned in values.yaml and push it to the repo. This will be done only if the utility has been changed.
- Now, we are using the tag of values.yaml, so no more need of TAG env variable in tests

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -9,16 +9,36 @@ on:
 
 jobs:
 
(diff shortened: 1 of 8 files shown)


madhurtrilio (approved):

madhurnawandar (approved):

udnay (approved):
LGTM

PR #156, opened 2021-08-19
Remove the manual steps while upgrading the cockroachdb helm chart

prafull01:
Currently, we are not deleting the SA, role and rolebinding from pre install hook and using those in cert rotation cronjob. These objects are not part of helm release and while upgrading user needs to delete those manually.

This PR removes those SA, role and rolebinding after pre jobs and for rotate certs a different sets is created which is  a part of helm chart.

### DIFF
--- a/cockroachdb/README.md
+++ b/cockroachdb/README.md
@@ -199,19 +199,7 @@ Exit the shell and delete the temporary pod:
 > \q
 ```
 
-If `tls.certs.selfSigner.enabled` was enabled for installing the cockroachdb then few resources need to be deleted first before upgrading.
-
-```shell
-kubectl delete sa <release-name>-self-signer
-kubectl delete role <release-name>-self-signer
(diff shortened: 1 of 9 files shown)


udnay (commented):
Changes look ok to me, can we add a test that checks to make sure there are no dangling SAs? Or a test that would have failed before this change?

prafull01:
@udnay After this change, the helm hook resources will be deleted in both `hook-succeeded,hook-failed` case, means hooks resources to be deleted in both the cases. Rest of resources are part of helm release and successful `helm uninstall` gives us guarantee that those resources be deleted on helm chart. 

The only resources which remains is the certificate secret which we create as a hook and we have written cleanup job which runs as a pre-delete. @abhishekdwivedi3060 has already started to write test cases for them.

I don't think we should test these SA, role and rolebinding deletion as these are guaranteed by helm to us.

PR #160, opened 2021-08-24
Add test-case for pre-delete e2e

abhishekdwivedi3060:
* Add test-case for pre-delete job which deletes the dangling secrets before the helm uninstall is done


udnay (approved):
LGTM

PR #168, opened 2021-09-13
Fix the job failure on merging the PRs without tag change

prafull01:
RCA:
 In case of tag not changed: `build/self-signer-utility.sh` returns exit code 1 which is treated as failure by github actions.

Fix:
 Instead of using the return code in the github actions, use the output returned by the script


udnay-crl (approved):
LGTM

PR #183, opened 2021-09-30
add ability to use extra volumes in statefulset

### LINKED ISSUE #182: allow to add extra volumes in cockroachdb pods
## Use case

We need to take backups of cockroachdb (like everybody spould) and export them to an external repository
For doing so we want to use tanzu/velero project.

## Analysis

Since there is no other volumes tahn datadir and certs there is not good candidate to dump the databases.
We could have been end up using datadir but this meens to sync also the real data files of cockroachdb which is both not needed and a ressource exausting

## Solution

We can allow to pass extra volumes to STS pods in order to have an easy way to backup our dbs
This can also be used for other use cases since this is quite a generic approach
Seems to be the way taken for a lot helm charts of the community, especially bitnami ones so it can be considered as a good practice

fredleger:
will fix #182

### DIFF
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -257,6 +257,9 @@ spec:
               mountPath: /cockroach/log-config
               readOnly: true
           {{- end }}
+          {{- with .Values.statefulset.extraVolumes.volumeMounts }}
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           livenessProbe:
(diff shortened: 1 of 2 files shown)


fredleger:
ready to merge on my side

fredleger:
any update on this one ? we do maintain a private fork of the helm chart just for it :-)

udnay:
This looks fine to me, can you please add some tests for this though.

fredleger:
one year happy bday !

PR #198, opened 2021-10-27
force new build of selfSigner image

JorritSalverda:


celiala:
> change line 3 of self-signer-utility.sh from
> ```tag=$(yq r ./cockroachdb/values.yaml 'tls.selfSigner.image.tag')```
> to 
> ```tag=$(yq e '.tls.selfSigner.image.tag' ./cockroachdb/values.yaml)```

Thanks for the suggestion!

For context, I'm not as familiar with the self-signer work. More came over here because we're seeing this error in the cockroachdb repository scan alert emails:

```
error scanning image gcr.io/cockroachlabs-helm-charts/cockroach-self-signer-cert:1.2: image not found (package cockroachdb:6.2.8)
```

After tweaking the make commands as per @JorritSalverda 's suggestion, running `make build-self-signer` worked locally for me (See #200)

@prafull01 @udnay @DuskEagle - I don't have write permissions to `gcr.io/cockroachlabs-helm-charts/cockroach-self-signer-cert`, but if the above looks good to you, perhaps you could push a 1.2 image via `make push-self-signer`, to resolve the error?

prafull01:
@JorritSalverda @celiala The helm-chart is fixed now with this PR: https://github.com/cockroachdb/helm-charts/pull/199 The issue was around the CI tag change job was not working properly. I have fixed it and now it should work for you..

PR #199, opened 2021-10-29
Add changes to fix the tag change CI jobs

prafull01:
- If tag is not changed in pre-build jobs, it should fail
- In post build job, removed the shell, it was not letting the job to write the output which is utilised by the next job


udnay (approved):

prafull01:
@JorritSalverda This PR fixed the CI to build the tag utility CI job. Now, the new build has been triggered and build successfully.

JorritSalverda:
Thanks @prafull01 I think we're almost there, except for chart `6.2.8` still using tag `1.2` which points to the same container image as `1.0` and `1.1` (see https://console.cloud.google.com/gcr/images/cockroachlabs-helm-charts/GLOBAL/cockroach-self-signer-cert).

When testing with values

```yaml
tls:
  selfSigner:
    image:
      tag: "1.3"
```

it does work as expected though. So I think we just need to have chart version `6.2.9` created from the latest values. Thanks for helping out.

JorritSalverda:
@prafull01 https://github.com/cockroachdb/helm-charts/pull/202 should create a new chart with the latest values

PR #206, opened 2021-11-21
Make init job annotations optional

scottcrossen:
Some kubernetes executors like ArgoCD will not deploy this chart correctly with the given annotations. I've found that these annotations need to be removed and replaced with:

```
argocd.argoproj.io/hook: Sync
argocd.argoproj.io/hook-delete-policy: HookFailed
```

This PR makes it possible to customize the annotations


pseudomuto:
This was addressed in #250

PR #208, opened 2021-11-22
Add extraPodSpec to statefulset and init job

oliverisaac:
This PR allows adding arbitrary values to the statefulset's pod spec. (As well as the init job.) This is needed to allow setting `shareProcessNamespace: true` to the pod spec to accommodate Hashicorp Vault injected dynamic certificates.

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -262,4 +262,7 @@ spec:
           {{- end }}
           {{- end }}
     {{- end }}
+    {{- with .Values.init.extraPodSpec -}}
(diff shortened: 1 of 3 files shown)


udnay:
Can you please add some tests for this feature and then we can take a look a it. 

The workflow you are trying to support is using vault to generate the certs instead of CRDBs self-signed version?

oliverisaac:
> The workflow you are trying to support is using vault to generate the certs instead of CRDBs self-signed version?

Yes, in general that's what I'm trying to support. Though, since I've made this PR, there are other fields I need to add to set cockroach up to not use `--insecure` while also not using the cert self signer.

PR #217, opened 2022-01-18
Add configuration options for certificates with the certificate manager

julburg:


cockroachdb/values.yaml
```diff
@@ -477,6 +477,14 @@ tls:
...
       kind: Issuer
       name: cockroachdb
+      # Duration of Client certificates in hours
+      clientCertDuration: 672h
+      # Expiry window of client certificates means a window before actual expiry in which client certs should be rotated.
+      clientCertExpiryWindow: 48h
+      # Duration of node certificates in hours
+      nodeCertDuration: 8760h
+      # Expiry window of node certificates means a window before actual expiry in which node certs should be rotated.
+      nodeCertExpiryWindow: `168h`
```
BuJo, line 487: The backticks seem suspect here.  I would revise this to use single quotes `'`.  The lines above would also benefit from quotation marks.

olle: Shouldn't this just be the value?

julburg: Yeah, thanks! It should only be the value!

PR #218, opened 2022-01-20
Make copy certs image in init container configurable

julburg:
* or else we run in a rate limiting problem


meridional:
Thanks for the contribution!

PR #222, opened 2022-02-10
Allow user to delete job pod on completion

EnriqueLop:
New parameter in init object that allow user to automatically delete init job pod when complete.

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -20,7 +20,11 @@ metadata:
   {{- end }}
   annotations:
     helm.sh/hook: post-install,post-upgrade
+    {{- if .Values.init.deleteJobOnCompletion }}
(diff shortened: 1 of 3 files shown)


cockroachdb/templates/job.init.yaml
```diff
@@ -20,7 +20,11 @@ metadata:
   {{- end }}
   annotations:
     helm.sh/hook: post-install,post-upgrade
+    {{- if .Values.init.deleteJobOnCompletion }}
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
+    {{- else }}
     helm.sh/hook-delete-policy: before-hook-creation
+    {{- end }}
```
abhishekdwivedi3060, line 27: We can keep this deletion behaviour as default one and add an option in values.yaml to disable this

abhishekdwivedi3060 (changes requested):

PR #227, opened 2022-02-24
Version update automation

rail:
Previously, in order to update the CockroachDB version, we used `sed` to
manipulate the files. This approach is error prone.

This patch adds a tool, that updates `Chart.yaml` and `values.yaml` files
with proper versions.

### DIFF
--- a/build/bump.go
+++ b/build/bump.go
@@ -1,0 +1,147 @@
+/*
+Copyright 2021 The Cockroach Authors
+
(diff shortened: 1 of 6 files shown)


rail:
`go run build/bump.go v21.2.7` gives me this diff https://gist.github.com/rail/29f8950e9c9854fef315ffe0c523aef0

celiala (approved):
This is pretty slick!! Thanks for adding this 🎉 .

LGTM!

PR #233, opened 2022-03-17
Add configurable pod termination grace period

pierreis:
The [CockroadhDB documentation](https://www.cockroachlabs.com/docs/stable/node-shutdown.html#termination-grace-period) implies that the pod termination grace period should be configurable depending on the cluster specifics.

The helm chart not only hardcodes the termination grace period but also sets it to a significantly shorter value than the recommended value in the documentation.

This merge request makes the termination grace period configurable.


prafull01:
This has been done by #502

PR #236, opened 2022-04-04
Fix issuer ref for the client and node certificates

julburg:
* the tls.certs.certManagerIssuer now also contains fields for the duration and expiry window


meridional:
Thanks for the fix.

PR #240, opened 2022-05-02
Update version to v21.2.10

absterr08:


cockroachdb/values.yaml
```diff
@@ -430,8 +430,6 @@ tls:
     # The name of this ServiceAccount to use.
     # If not set and `create` is `true`, then a name is auto-generated.
     name: ""
-  copyCerts:
-    image: busybox
```
rail, line 434: Hmmm. Why do we remove these?

cockroachdb/values.yaml
```diff
@@ -480,14 +478,6 @@ tls:
...
       kind: Issuer
       name: cockroachdb
-      # Duration of Client certificates in hours
-      clientCertDuration: 672h
-      # Expiry window of client certificates means a window before actual expiry in which client certs should be rotated.
-      clientCertExpiryWindow: 48h
-      # Duration of node certificates in hours
-      nodeCertDuration: 8760h
-      # Expiry window of node certificates means a window before actual expiry in which node certs should be rotated.
-      nodeCertExpiryWindow: 168h
```
rail, line 490: And these?

absterr08: I'm not sure 😅 these were generated based on the release script. Should they be left in?

rail: Let me double check. The script shouldn't be touching those...

rail:
https://github.com/cockroachdb/helm-charts/pull/241 is to address the issues.

PR #241, opened 2022-05-03
Update templates and bump to v21.2.10

rail:
In #217 some new configuration parameters were introduced, but we forgot
to update the template used to generate the actual files.

This PR updates the templates in order to reflect the current state.


absterr08 (approved):
LGTM - thanks Rail!

PR #243, opened 2022-05-23
Update version to v22.1.0

cameronnunez:


cockroachdb/Chart.yaml
```diff
@@ -2,8 +2,8 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 7.0.3
-appVersion: 21.2.10
+version: 7.0.4
```
celiala, line 5: There is a bug in our script:

Can you bump `version` to `8.0.0`?

(i.e. a major version bump for `appVersion` should also mean a major version bump for `version`)

Thank you!

cameronnunez: Done!

PR #247, opened 2022-05-31
Chart major version should be bumped with new CRDB major releases

rail:


celiala (approved):
LGTM! Thanks for fixing this! (sorry for the review lag 😅 )

Issue #149, opened 2021-08-02
Cockroachdb fails to init with argocd


fredleger:
Hi folks,

I guess we hit an issue quite similar to #69 with argocd

With argocd helm hooks are converted automatically to argocd hooks.
The init job does include annotations with `post-install` and `post-upgrade` stages. They will be converted in `postSync` in argocd (source: https://argoproj.github.io/argo-cd/user-guide/helm/#helm-hooks).

In fact the job will never been triggered because the sync stage will never be over untill all pods are `healthy` which can't occurs without init. inifinite loop :-)

One easy way to fix that is to allow customization of Job annotations (`Job.spec.annotations` not `Job.spec.template.annotations` which are already supported with `init.annotations`). This will allow to add an annotation for Argocd like `argocd.argoproj.io/hook: Sync` which i guess could do the job.

fredleger:
I can confirm than injecting the job with the abvove annotations initialize the cluster as expected

wodka:
would be really nice to add this to the general readme!

```
init:
  jobAnnotations:
    argocd.argoproj.io/hook: Sync
```

Issue #249, opened 2022-06-08
Invalid order when cert-manager is used with TLS enabled


diogokiss:
**Version**

CockroachDB Helm chart: 7.0.1 (https://charts.cockroachdb.com/)
(I strongly suspect that the same issue hits the latest version too)

**CockroachDB Helm chart configuration**
```
cockroachdb:
  tls:
    enabled: true
    certs:
      selfSigner:
        enabled: false
      certManager: true
      useCertManagerV1CRDs: true
      tlsSecret: true
      certManagerIssuer:
        kind: ClusterIssuer
```

**ClusterIssuer**

```
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: letsencrypt-prod
spec:
  acme:
    email: redacted@domain.com
    privateKeySecretRef:
      name: letsencrypt-key-prod
    server: https://acme-v02.api.letsencrypt.org/directory
    solvers:
    - dns01:
        cloudDNS:
          project: <my-gcp-project>
```

**Error**

```
Events:
  Type     Reason           Age   From                                          Message
  ----     ------           ----  ----                                          -------
  Warning  InvalidOrder     13m   cert-manager-certificaterequests-issuer-acme  The CSR PEM requests a commonName that is not present in the list of dnsNames or ipAddresses. If a commonName is set, ACME requires that the value is also present in the list of dnsNames or ipAddresses: "root" does not exist in [] or []
  Normal   cert-manager.io  13m   cert-manager-certificaterequests-approver     Certificate request has been approved by cert-manager.io
```

**Related to**
https://github.com/cert-manager/cert-manager/issues/2370#issuecomment-554372660

**Potential solutions**

1. Add the `commonName` to the `dnsNames` list
2. Make the `commonName` customizable in the Helm chart (I'm not sure whether this would mess with any validation in the application itself though)
3. Both alternatives above

This issue similarly affects the order generated for the `node` certificate too:
```
Events:
  Type     Reason           Age    From                                          Message
  ----     ------           ----   ----                                          -------
[...]

Issue #230, opened 2022-03-08
Include Deprecation Notice


jaredpetersen:
I started using this Helm chart before realizing that it's deprecated according to the [documentation](https://www.cockroachlabs.com/docs/stable/deploy-cockroachdb-with-kubernetes.html?filters=helm#step-2-start-cockroachdb):

> **Warning:**
> The CockroachDB Helm chart is compatible with Kubernetes versions 1.22 and earlier (the latest version as of this writing). However, no new feature development is currently planned. If you are experiencing issues with a Helm deployment on production, contact our [Support team](https://support.cockroachlabs.com/).

There should be a notice on this repository too so that people know to switch to the Operator for any new projects.

PR #250, opened 2022-06-08
Add support for custom annotations to the Init Job 

diogokiss:
In some cases, there is a need for adding custom annotations to the Init job.
For example, when using ArgoCD, we need to instruct ArgoCD to start the job
still during the Sync phase and not as a PostSync hook. Otherwise, the job never
starts. And this is done by adding the following annotation to the job.
```
argocd.argoproj.io/hook: Sync
```

This is to continue the work started in https://github.com/cockroachdb/helm-charts/pull/246 and https://github.com/cockroachdb/helm-charts/pull/206
to address https://github.com/cockroachdb/helm-charts/issues/149

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 8.0.1
+version: 8.1.0
 appVersion: 22.1.1
 description: CockroachDB is a scalable, survivable, strongly-consistent SQL database.
 icon: https://raw.githubusercontent.com/cockroachdb/cockroach/master/docs/media/cockroach_db.png
--- a/cockroachdb/values.yaml
+++ b/cockroachdb/values.yaml
@@ -362,6 +362,9 @@ init:
   labels:
     app.kubernetes.io/component: init
 
+  # Additional annotations to apply to this Job.
+  jobAnnotations: {}
+
   # Additional annotations to apply to the Pod of this Job.
   annotations: {}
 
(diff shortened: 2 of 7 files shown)


diogokiss:
@pseudomuto and @rail, could you maybe take a look at this PR, please? 😬

cockroachdb/Chart.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 8.0.1
+version: 8.0.2
```
pseudomuto, line 5: Since we're adding a new feature I think we should make this new version `8.1.0`.

pseudomuto (approved):
Aside from the version this LGTM 👍 Nice work!

cockroachdb/values.yaml
```diff
@@ -362,6 +362,9 @@ init:
   labels:
     app.kubernetes.io/component: init
 
+  # Additional annotations to apply to this Job.
+  jobAnnotations: {}
+
```
rail, line 367: At the top of this file it says 
```
# Generated file, DO NOT EDIT. Source: build/templates/values.yaml
```
;)
If you don't update it, these changes will be overwritten.

rail (changes requested):
Would you mind adding an example usage somewhere? Maybe to the Readme template or as comments next to the new section. What do you think?

cockroachdb/values.yaml
diogokiss: Done. ✅

cockroachdb/Chart.yaml
diogokiss: Done. ✅

diogokiss:
@rail and @pseudomuto, I've addressed and force-pushed the requested changes. 🤝

diogokiss:
On another note, I saw [this](https://github.com/cockroachdb/helm-charts/issues/230). What is the recommended way of installing CockroachDB in a K8s cluster? Is it via Helm or Operator? Just for clarification. Also, I have [another issue](https://github.com/cockroachdb/helm-charts/issues/249) opened that seems not to affect the Operator. So, if I'm supposed to use the Operator, I'll not even bother to wait for it to be resolved.

rail (approved):
Thank you very much!

Issue #147, opened 2021-08-01
Cockroachdb installation with TLS enabled through certificate manager is failing


prafull01:
I have tried cockroachdb installation with cert manager and the installation doesn't succeed and goes into `crashloopbackoff` stage.

Here is my certificate issuer:
```
prafull@EMPID18004:helm-charts$ kubectl get issuers.cert-manager.io 
NAME                      READY   AGE
cockroachdb-cert-issuer   True    17h
```

Installed the helm chart using following attributes:
```
tls:
-  enabled: false
+  enabled: true
   serviceAccount:
     # Specifies whether this ServiceAccount should be created.
     create: true
@@ -395,14 +395,14 @@ tls:
     # TLS secrets are created by cert-mananger, for example.
     tlsSecret: false
     # Use cert-manager to issue certificates for mTLS.
-    certManager: false
+    certManager: true
     # Specify an Issuer or a ClusterIssuer to use, when issuing
     # node and client certificates. The values correspond to the
     # issuerRef specified in the certificate.
     certManagerIssuer:
       group: cert-manager.io
       kind: Issuer
-      name: cockroachdb
+      name: cockroachdb-cert-issuer
```

After installing the cockroachdb :
```
prafull@EMPID18004:helm-charts$ kgp
NAME                          READY   STATUS             RESTARTS   AGE
crdb-cockroachdb-0            0/1     CrashLoopBackOff   2          57s
crdb-cockroachdb-1            0/1     CrashLoopBackOff   2          57s
crdb-cockroachdb-2            0/1     Error              2          57s
crdb-cockroachdb-init-srbft   1/1     Running            0          56s
```

The secrets are properly created by the cert manager:
```
prafull@EMPID18004:helm-charts$ kubectl get secrets 
NAME                           TYPE                                  DATA   AGE
ca-secret-auth                 kubernetes.io/tls                     2      17h
cockroachdb-node               kubernetes.io/tls                     3      105s
cockroachdb-root               kubernetes.io/tls                     3      105s
crdb-cockroachdb-token-8dqd5   kubernetes.io/service-account-token   3      107s
default-token-44bxk            kubernetes.io/service-account-token   3      24h
sh.helm.release.v1.crdb.v1     helm.sh/release.v1                    1      108s
```

PR #253, opened 2022-06-29
Update statefulset.yaml

alleny87:
When Certs are provided via CertManager, the /cockroach/certs volume mount needs to be added to allow the secrets to be mounted into the pod


prafull01:
Fixed with #147 . Closing this.

PR #256, opened 2022-07-11
Add isito quitquitquit command to init job

DoGab:
Make it possible to use Istio. The init job pod never completes if istio is deployed. Therefore we need to wait until the envoy proxy is ready and after everything is done we need to tell the proxy to quit.


DoGab:
What is needed to get this merged?

PR #262, opened 2022-09-08
avoid using .Release.Namespace

ooraini:
It's difficult to use any sort of [post-render](https://helm.sh/docs/topics/advanced/#post-rendering) such as Kustomize when {{ .Release.Namespace }} is used inside cmds, args or scripts.

Usage of {{ .Release.Namespace }} can be replaced with:
```
    - name: NAMESPACE
      valueFrom:
        fieldRef:
          fieldPath: metadata.namespace
```

In addition to the chart templates, the cleanup command takes `namespace` as a flag. Which is not consistent with other command and forces the use of bash for substitution.


prafull01:
Hi @ooraini Thank you for your contribution on this. Would you like to revive this PR. Can you please rebase it with latest changes.

PR #265, opened 2022-09-23
Configure prometheus scrape options

tomellis91:
This PR is being raised to provide the ability to disable automatic monitoring by prometheus. 
- Default option is still set to true
- Added a conditional statement to allow the option to ignore setting the prometheus annotations


meridional:
LGTM. Thanks for the contribution!

PR #267, opened 2022-09-26
Add option to set the ingress class name

mtneug:
The ingressClassName configuration option should be used instead of custom ingress dependent annotations. This adds a new value to the ingress configuration to set this option.

### DIFF
--- a/build/templates/values.yaml
+++ b/build/templates/values.yaml
@@ -440,6 +440,7 @@ ingress:
   annotations: {}
   #   kubernetes.io/ingress.class: nginx
   #   cert-manager.io/cluster-issuer: letsencrypt
+  className: ""
   paths: [/]
   hosts: []
   # - cockroachlabs.com
--- a/cockroachdb/templates/ingress.yaml
+++ b/cockroachdb/templates/ingress.yaml
@@ -32,6 +32,9 @@ metadata:
 {{- toYaml .Values.ingress.labels | nindent 4 }}
 {{- end }}
 spec:
+{{- if .Values.ingress.className }}
+  ingressClassName: {{ .Values.ingress.className }}
+{{- end }}
   rules:
   {{- if .Values.ingress.hosts }}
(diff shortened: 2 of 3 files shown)


zadjadr:
It would make sense to also check beforehand if the ingress resource can handle `ingressClassName` like well known other helm repos do (e.g. Grafanalabs: https://github.com/grafana/helm-charts/blob/316f4a28aed5aa35819fbc11673ab76e1a95cede/charts/tempo/templates/_helpers.tpl#L93-L105)

mtneug:
IMO at this point I would move on to only support `networking.k8s.io/v1`. This version was GA in Kubernetes 1.19 (released mid 2020).

jnt2007:
Hi. Why is it stuck?

mtneug:
> Hi. Why is it stuck?

No clue. I have used an internal fork for years now, mostly because of this.

jnt2007:
Any suggestion on how we can push it forward?

prafull01 (commented):
@mtneug Can you please do following:
1. Rebase the PR with latest master.
2. Fix the lint tests: you need to run: `go run build/build.go generate` to generate the build templates.

jnt2007:
The current documentation says that the [Helm chart supports Kubernetes 1.8 ](https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/README.md?plain=1#L15)and newer. The `ingressClassName` attribute [has been supported since Kubernetes 1.18](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation).

If I have a vote, I also think that verification of the k8s version is overkill here, because k8s 1.8 is the ancient version.

mtneug:
@prafull01 thanks for confirming.

Hopefully, the current patch will make it through CI.

prafull01 (approved):

mtneug:
Thanks @prafull01 for the quick review.

prafull01:
Thanks @mtneug for your contribution. :)

PR #272, opened 2022-09-29
allow to provide the users password in an existing secret

glehmann:
This is useful to not pass the password in the values, for example
when using this chart as a dependency of another one or when working
with sealed secrets.


glehmann:
Just closing my outdated pull requests.

PR #273, opened 2022-09-29
Adding security context for Statefulsets

grohan2002:
The changes in this PR contain Security Context for node Statefulsets.

It covers -
container spec.
```securityContext:
  allowPrivilegeEscalation: false,
  capabilities:
    drop":
    - "ALL"
  privileged: false,
  readOnlyRootFilesystem: true```
PodSpec’s SecurityContext
```fsGroup: 1000
runAsGroup: 1000
runAsNonRoot:	true
runAsUser	1000```

### DIFF
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -284,8 +284,16 @@ spec:
             periodSeconds: 5
             failureThreshold: 2
           {{- end }}
-        {{- with .Values.statefulset.securityContext }}
-          securityContext: {{- toYaml . | nindent 12 }}
+        {{- if eq (include "cockroachdb.securityContext.versionValidation" .) "true" }}
+        {{- if .Values.statefulset.securityContext.enabled }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
+        {{- end }}
         {{- end }}
         {{- with .Values.statefulset.resources }}
           resources: {{- toYaml . | nindent 12 }}
@@ -342,8 +350,14 @@ spec:
           secret:
             secretName: {{ template "cockroachdb.fullname" . }}-log-config
       {{- end }}
-      {{- with .Values.securityContext }}
-      securityContext: {{- toYaml . | nindent 8 }}
+      {{- if eq (include "cockroachdb.securityContext.versionValidation" .) "true" }}
+      {{- if and .Values.securityContext.enabled }}
+      securityContext:
+        fsGroup: 1000
+        runAsGroup: 1000
+        runAsUser: 1000
+        runAsNonRoot: true
+      {{- end }}
       {{- end }}
 {{- if .Values.storage.persistentVolume.enabled }}
   volumeClaimTemplates:
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -284,6 +284,9 @@ spec:
             periodSeconds: 5
             failureThreshold: 2
           {{- end }}
+        {{- with .Values.statefulset.securityContext }}
+          securityContext: {{- toYaml . | nindent 12 }}
+        {{- end }}
         {{- with .Values.statefulset.resources }}
           resources: {{- toYaml . | nindent 12 }}
         {{- end }}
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by yecs1999** [comment]
Hey, it looks good so far, but have you tested it with these instructions yet? https://www.cockroachlabs.com/docs/v22.1/deploy-cockroachdb-with-kubernetes.html If not I can test it.

**Comment by grohan2002** [comment]
I did test it. Do you want me to post any specific results I can do it tomorrow morning my time

**Comment by yecs1999** [comment]
Yes that would be good, if you can post them tomorrow. You ran commands like `id` to verify it was using the non root user right? also check the filesystem is read only.

**Comment by grohan2002** [comment]
@yecs1999 I did setup 2 CDB clusters using helm charts -

One with the Security Context changes and one without it. As evident the user is non root vs root.

```rohan@Rohans-iMac cockroachdb % kubectl exec -it cdb-security-cockroachdb-0  bash
bash-4.4$ 
bash-4.4$ id
uid=1000 gid=1000 groups=1000
bash-4.4$ exit
exit
rohan@Rohans-iMac cockroachdb % kubectl exec -it my-release-cockroachdb-0 bash
[root@my-release-cockroachdb-0 cockroach]# id
uid=0(root) gid=0(root) groups=0(root)
```

yecs1999 (approved):
Thanks for testing, LGTM

cockroachdb/templates/statefulset.yaml
```diff
@@ -339,6 +342,9 @@ spec:
           secret:
             secretName: {{ template "cockroachdb.fullname" . }}-log-config
       {{- end }}
+      {{- with .Values.securityContext }}
+      securityContext: {{- toYaml . | nindent 8 }}
+      {{- end }}
```
catj-cockroach, line 347: Rather than passing in the raw security context, could we make this work like the `Values.storage.persistentVolume.enabled` flag? (Likewise for the StatefulSet).

We should also add a comment or a check for the CockroachDB version that advises people to disable it if running an older version of CockroachDB. @yecs1999 could you confirm the minimum versions we'd need to require?

```suggestion
      {{- if .Values.securityContext.enabled }}
      securityContext:
        fsGroup: 1000
        runAsGroup: 1000
        runAsUser: 1000
        runAsNonRoot: true
      {{- end }}
```

catj-cockroach (changes requested):

cockroachdb/templates/statefulset.yaml
yecs1999: Good catch. @grohan2002 do you know if you can test this with the version tag [here](https://github.com/cockroachdb/helm-charts/blob/d7e28f049f381c76879d9394eb26123196682a56/cockroachdb/values.yaml#L2-L4) changed to 22.1.1 and 21.2.12? Both of these versions are below the minimum required. If there is an error there then we need to have some kind of switch so any versions that are unsupported (below 21.2.13 or below 22.1.2) will not have this securitycontext field populated.

grohan2002:
We are adding the entire securityContext in the statefulset.yaml so we can't introduce the custom field enabled in the securityContext section. So I suggest the field would be like this in values.yaml
```
security:
  enabled: true
  securityContext:
    allowPrivilegeEscalation: false
    capabilities:
      drop:
        - ALL
    privileged: false
    readOnlyRootFilesystem: true
```
We will check .Values.security.enabled before adding this value.
We will also add a _helpers.tpl validation function which will set the security.enabled as false if the version is less than 22.1.2 and 21.2.13
Do comment if this suffices @catj-cockroach @yecs1999

yecs1999:
@grohan2002 I think this is sufficient, as long as we have the way to switch the securitycontext on and off by version.

catj-cockroach:
@grohan2002 Oh, sorry I mean rather than having the `securityContext` value in Values.yml, having the entire `securityContext` inline in the statefulset and only having the `enabled` value in Values.yml. I'd argue it doesn't really make sense to have the `securityContext` definitions in Values.yml since it's unlikely that someone would be tweaking it unless they want to do something outside of the norm (in which case they shouldn't be using the helm chart).

yecs1999 (changes requested):

catj-cockroach (approved):

HaveFun83:
Great feature. Any ETA when this is available?

mmack:
Any news on this? Please merge :)

prafull01 (approved):

prafull01:
We have two approvals on this PR, Merging this.

rail:
I think we missed https://github.com/cockroachdb/helm-charts/blob/master/build/templates/values.yaml

Issue #270, opened 2022-09-27
policy/v1beta1 PodDisruptionBudget Deprecated in v1.21+, unavailable in v1.25+


sergeyshaykhullin:
Helm CRDB version: 8.1.6

I am on 1.25 and this check https://github.com/cockroachdb/helm-charts/blob/7d35ff82c2e291496b21b536ca7a348584195a81/cockroachdb/templates/poddisruptionbudget.yaml#L2 still select beta pdb 

```
Error: UPGRADE FAILED: unable to recognize "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
```

sergeyshaykhullin:
Not fixed

```
+ helm_v3 upgrade --namespace crdb --repo https://charts.cockroachdb.com --version 10.0.7 crdb cockroachdb --values /config/values-01_HelmChart.yaml
Error: UPGRADE FAILED: unable to build kubernetes objects from current release manifest: resource mapping not found for name: "crdb-budget" namespace: "crdb" from "": no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"
ensure CRDs are installed first
```

prafull11:
It is working on k8s 1.25. 
<img width="1440" alt="Screenshot 2023-03-31 at 12 05 17 AM" src="https://user-images.githubusercontent.com/56869150/228932247-023fc6d0-afc7-4098-881b-b4817575cc3a.png">

Next Steps:
1. Please check the helm version on your system?
2. Run the `helm install` command with --debug mode and see if the PDB yaml is generated correctly.

sergeyshaykhullin:
Hmm, we are on k3s v1.26.3

I know it should use policy/v1, but it is not...

Maybe bacause of mapkubeapis
```
helm_v3 mapkubeapis crdb --namespace crdb
Already installed crdb
2023/03/30 19:33:17 Release 'crdb' will be checked for deprecated or removed Kubernetes APIs and will be updated if necessary to supported API versions.
2023/03/30 19:33:17 Get release 'crdb' latest version.
2023/03/30 19:33:17 Check release 'crdb' for deprecated or removed APIs...
2023/03/30 19:33:17 Finished checking release 'crdb' for deprecated or removed APIs.
2023/03/30 19:33:17 Release 'crdb' has no deprecated or removed APIs.
2023/03/30 19:33:17 Map of release 'crdb' deprecated or removed APIs to supported versions, completed successfully.
```

[...]

PR #275, opened 2022-09-29
Fix pdb api version for k8s 1.21+

sergeyshaykhullin:
Current if condition is wrong and chart is not installable for k8s 1.25+ because of policy/v1beta1 selection

- [ ] Fix https://github.com/cockroachdb/helm-charts/issues/270


sergeyshaykhullin:
@rhu713 @celiala Can you please merge this quick fix?

sergeyshaykhullin:
Same check from otel collector helm chart

https://github.com/open-telemetry/opentelemetry-helm-charts/blob/fe8ee564b0c6c76f14e3af20daf318ae28855dce/charts/opentelemetry-collector/templates/_helpers.tpl#L153

meyfa:
Pinging to hopefully get some maintainer attention - we need this fix desperately. We currently cannot deploy CockroachDB in our cluster due to this.

Quick note: The same problem exists with CronJob resources, as well.

rail (approved):
Thank you for the fix!

PR #276, opened 2022-10-06
Update terminationGracePeriodSeconds to 300 seconds

mdlinville:
A value of 0 is strongly recommended against in the Kubernetes docs (https://github.com/kubernetes/website/pull/1731/files#diff-da05a631829c13713cb0f46fc58f809d75db6d624c62a292fc787faf441e612cR38). The CockroachDB public operator uses a value of 300 seconds (https://github.com/cockroachdb/cockroach-operator/blob/master/pkg/resource/statefulset.go#L52).

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -49,7 +49,7 @@ spec:
     {{- end }}
     {{- end }}
       restartPolicy: OnFailure
-      terminationGracePeriodSeconds: 0
+      terminationGracePeriodSeconds: 300
     {{- if or .Values.image.credentials (and .Values.tls.enabled .Values.tls.selfSigner.image.credentials (not .Values.tls.certs.provided) (not .Values.tls.certs.certManager)) }}
       imagePullSecrets:
       {{- if .Values.image.credentials }}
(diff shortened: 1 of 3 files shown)


noguchitoshi:
It looks like we also set this value in these locations:
- [job.init.yaml](https://github.com/cockroachdb/helm-charts/blob/193ed8f59f81959fbe4b73478f2a9ea1306ed94c/cockroachdb/templates/job.init.yaml#L41)
- [statefulset.yaml](https://github.com/cockroachdb/helm-charts/blob/193ed8f59f81959fbe4b73478f2a9ea1306ed94c/cockroachdb/templates/statefulset.yaml#L141)

@pseudomuto, I would imagine we'd want to keep things consistent at 300?

pseudomuto (approved):
Yeah, the statefulset is the one we really care about. Once we update that, this LGTM 👍 Thanks for doing this @mdlinville!

mdlinville:
@pseudomuto Is there any convention about release notes in this repo? Updating defaults seems like it might need a release note.

prafull01 (approved):

prafull01:
@mdlinville Can you please rebase your branch with master so that github workflows can run. Merging is allowed only after the github workflows passes.

mdlinville:
> @mdlinville Can you please rebase your branch with master so that github workflows can run. Merging is allowed only after the github workflows passes.

ANy idea aBout the build failure after rebase?

PR #277, opened 2022-10-17
update to v22.1.9

celiala:


cockroachdb/README.md
```diff
@@ -362,7 +362,6 @@ For details see the [`values.yaml`](values.yaml) file.
 | `ingress.hosts`                                           | CockroachDB Ingress hostnames                                   | `[]`                                                  |
 | `ingress.tls[0].hosts`                                    | CockroachDB Ingress tls hostnames                               | `nil`                                                 |
 | `ingress.tls[0].secretName`                               | CockroachDB Ingress tls secret name                             | `nil`                                                 |
-| `prometheus.enabled`                                      | Enable automatic monitoring of all instances when Prometheus is running | `true`                                        |
```
rail, line 365: Hmm, probably someone instead of changing the template updated the generated file... I wonder if we should add a lint check or something to prevent changing the generated file, except some allowed lines 🤔

celiala: TFTR!

Updated PR to incorporate changes from https://github.com/cockroachdb/helm-charts/pull/265 into the build template files. PR diff is now just the version bump changes.

>I wonder if we should add a lint check or something to prevent changing the generated file, except some allowed lines 🤔

Good idea -- created ticket: https://cockroachlabs.atlassian.net/browse/RE-303

PR #278, opened 2022-10-18
introduce configurable resource limits for the copy-certs container

DrFelder:
Resources for copy-certs has to be configurable as OPA gatekeepers etc. will not only require resource limits for pods, but all containers

### DIFF
--- a/build/templates/values.yaml
+++ b/build/templates/values.yaml
@@ -432,6 +432,18 @@ tls:
     name: ""
   copyCerts:
     image: busybox
+    # The copy-certs container copies provided certificates. It finishes
+    # quickly and doesn't continue to consume resources in the Kubernetes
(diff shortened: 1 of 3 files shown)


cockroachdb/values.yaml
```diff
@@ -426,6 +426,19 @@ init:
     #       # https://www.cockroachlabs.com/docs/stable/create-schedule-for-backup.html#schedule-options
     #       options: [first_run = 'now']
 
+copyCerts:
```
rail, line 429: Thank you for your contribution!
Can you edit `build/templates/values.yaml` instead of this file - the contents of this file will be replaced (it's generated).

DrFelder: done

I also moved the config to the tls.copyCerts section; the image of the container is already configured there.

rail (approved):

PR #281, opened 2022-11-15
bump version to v22.1.11

celiala:


rail (approved):

cockroachdb/values.yaml
```diff
@@ -438,6 +438,18 @@ tls:
     name: ""
   copyCerts:
     image: busybox
+    # The copy-certs container copies provided certificates. It finishes
```
celiala, line 441: template is out of sync with generated file again. lemme fix and update

PR #282, opened 2022-11-15
Release template linter

celiala:
This PR is a linter that fails if generated output from template produces a diff.

I can't quite figure out how to do this yet via github actions, but this is what I'm hoping to do:

In a github action:
- run `go run build/build.go generate`
- fail if the diff is not empty

research so far:
- I spent a little bit of time trying to find something like this in https://github.com/marketplace - haven't gotten anything concrete[1] yet.
	- [1] part of this could be me not fully understanding the format / features available for github actions ;)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -146,3 +146,24 @@ jobs:
 
       - name: Run E2E Test
         run: make test/e2e/rotate
+
+  lint-templates:
+    name: Lint release templates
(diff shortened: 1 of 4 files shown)


celiala:
@rail - curious what you think of this as an idea for a linter to confirm that generated files match their templates

rail:
I like the approach! Easy and straight forward! A nit. Maybe we should use the flag module instead of `os.Args`? 🤔

PR #284, opened 2022-11-18
Document resource limits for the copy-certs container

miwig:
Add documentation for the changes introduced in #278


rail (approved):
Thank you!

PR #295, opened 2023-02-16
bump version to 22.2.5

celiala:


cockroachdb/Chart.yaml
```diff
@@ -2,8 +2,8 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 10.0.3
-appVersion: 22.2.3
+version: 10.0.5
```
celiala, line 5: I intentionally bumped version from 10.0.3 to 10.0.5 (in case there were already any 10.0.4 in the wild)

cockroachdb/values.yaml
```diff
@@ -437,7 +437,7 @@ tls:
     # If not set and `create` is `true`, then a name is auto-generated.
     name: ""
     # Additional serviceAccount annotations (e.g. for attaching AWS IAM roles to pods)
-    annotations: {}
```
celiala, line 440: there's a discrepancy between template and generated yaml. fixing now.

rail (approved):

PR #297, opened 2023-03-07
Fixed Network Policy Template

PascalTurbo:
This Bugfix was originally created by @travigd and is discussed here: https://github.com/cockroachdb/helm-charts/pull/25

I've only bumped the chart version so the fix should be mergable.

### DIFF
--- a/cockroachdb/Chart.yaml
+++ b/cockroachdb/Chart.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 10.0.6
+version: 10.0.7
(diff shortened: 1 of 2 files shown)


hermannolafs:
Any news on getting this into a release? This would benefit our team greatly

cockroachdb/Chart.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 10.0.6
+version: 10.0.7
```
Spacebjorn, line 5:
```suggestion
version: 10.0.6
```

Issue #283, opened 2022-11-18
Error "Unable to create ClusterRole" if two releases with same releasename in different namespaces


amakarov70:
I insatll two releases with same releasename "cockroachdb" in two namespace "cockroachdb" and "cockroachdb2"
After second helm install:
_Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "cockroachdb" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "cockroachdb2": current value is "cockroachdb"_

In chart ClusterRole have namespace and differen namespaces, but ClusterRole is a cluster-level and not belong to namespace.
Maybe need change name to: **{{ template "cockroachdb.fullname" . }}-{{.Release.NameSpace}}**

prafull01:
Fixed by #301

PR #301, opened 2023-03-21
Add the release namespace in cluster scope resources

### LINKED ISSUE #300: Helm install fails if same release name is used in two different namespaces
Right now if you install cockroachdb in two different namespace with same release name. The first install is successful.

```
prafulll@18004 helm-charts % helm install crdb ./cockroachdb -n crdb-1
NAME: crdb
LAST DEPLOYED: Tue Mar 21 22:34:23 2023
NAMESPACE: crdb-1
STATUS: deployed
REVISION: 1
NOTES:
CockroachDB can be accessed via port 26257 at the
following DNS name from within your cluster:

crdb-cockroachdb-public.crdb-1.svc.cluster.local

Because CockroachDB supports the PostgreSQL wire protocol, you can connect to
the cluster using any available PostgreSQL client.

Note that because the cluster is running in secure mode, any client application
that you attempt to connect will either need to have a valid client certificate
or a valid username and password.

Finally, to open up the CockroachDB admin UI, you can port-forward from your
local machine into one of the instances in the cluster:

    kubectl port-forward crdb-cockroachdb-0 8080

Then you can access the admin UI at https://localhost:8080/ in your web browser.

For more information on using CockroachDB, please see the project's docs at:
https://www.cockroachlabs.com/docs/
```

Second install fails with following error:
```
prafulll@18004 helm-charts % helm install crdb ./cockroachdb -n crdb-2
Error: INSTALLATION FAILED: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "crdb-cockroachdb" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "crdb-2": current value is "crdb-1"
```

prafull01:
Fixes #300 #283 

Added namespace name in cluster scope resources so that the name doesn't conflict when the same release name is used.

### DIFF
--- a/cockroachdb/templates/_helpers.tpl
+++ b/cockroachdb/templates/_helpers.tpl
@@ -24,6 +24,24 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 
 {{/*
+Create a default fully qualified app name for cluster scope resource.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name with release namespace appended at the end.
+*/}}
+{{- define "cockroachdb.clusterfullname" -}}
+{{- if .Values.fullnameOverride -}}
+    {{- printf "%s-%s" .Values.fullnameOverride .Release.Namespace | trunc 56 | trimSuffix "-" -}}
+{{- else -}}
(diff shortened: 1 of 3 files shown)


rail (approved):
LGTM, :shipit:

PR #305, opened 2023-03-28
Add securityContext to template

rail:
In #273 we updated the generated file, but forgot to update the template.


rail:
cherry-picked in https://github.com/cockroachdb/helm-charts/pull/282

PR #307, opened 2023-04-25
Fix `securityContext.versionValidation` for custom CockroachDB images

tyranron:
## Synopsis

There is a `cockroachdb.securityContext.versionValidation` helper in `_helpers.tpl` pre-checking whether `securityContext.enabled` value should have an effect or not. The problem with this helper that it assumes `Values.image.repository` always to be a `cockroachdb/cockroach`, or another image being versioned exactly as the upstream one. 

We do use this Helm chart with our custom CockroachDB images, versioned in a very different way (using Git branch name as an image tag), and as the result this Helm chart stopped working for us with the following error:
```
Error: template: backend/charts/cockroachdb/templates/statefulset.yaml:290:19: executing "backend/charts/cockroachdb/templates/statefulset.yaml" at <include "cockroachdb.securityContext.versionValidation" .>: error calling include: template: backend/charts/cockroachdb/templates/_helpers.tpl:277:7: executing "cockroachdb.securityContext.versionValidation" at <semverCompare ">=22.1.2" .Values.image.tag>: error calling semverCompare: Invalid Semantic Version
```

## Solution

1. I've added a check whether the image is custom or the default one to `cockroachdb.securityContext.versionValidation` helper. Now it doesn't apply this check for custom images, so they can be used with this Helm chart again. And doesn't block using `securityContext.enabled` for custom images.

2. Also, I've removed using `cockroachdb.securityContext.versionValidation` helper for `.Values.tls.selfSigner.image`, since the Docker image used there is not connected with the `cockroachdb/cockroach` image versioning in any way.


cockroachdb/templates/_helpers.tpl
```diff
@@ -274,13 +274,18 @@ Validate that if user enabled tls, then either self-signed certificates or certi
 {{- end -}}
 
 {{- define "cockroachdb.securityContext.versionValidation" }}
+{{/* Allow using `securityContext` for custom images. */}}
+{{- if ne "cockroachdb/cockroach" .Values.image.repository -}}
+    {{ print true }}
+{{- else -}}
```
prafull01, line 280: The only issue with the approach is that any custom image built with version less than `21.2.13` will have the securityContext enabled and the cockroachdb pods will not come up.

Instead of returning true, we return false from here in custom images case. Also, we can have a flag `force` in `securityContext` section, if enabled  we allow the securityContext insertion in all cases, so custom image user's can enable the `force` flag to insert the security context. This `force` flag is only needed in statefulset security context.

@rail WDYT?

prafull01 (commented):
@tyranron Changes looks good so far 👍 . I have just one item to discuss and I have left a comment on it for further discussion.

cockroachdb/templates/_helpers.tpl
rail: Sounds sane to me. I'd rather have an explicit directive to use `securityContext` rather deriving it from the repo name.
Thank you for your contribution @tyranron!

tyranron: @prafull01 on my opinion, the described way is not better, because:
- Custom image of version less than `21.2.13` is more rare situation than a custom images built on top of latest CockroachDB version. Furthermore, as the time goes by, they will become even more and more rare.
- So, in my opinion, it's better to make a minority of users to opt-out `securityContext` explicitly, rather than making a majority of users to opt-in it explicitly.
- Also, because it's easy for the majority of users to skip using `securityContext` without wanting so. The thing is too subtle. You have your deployment being configured, then you just change the image repository and voilà - the `securityContext` is subtly disabled, no errors are printed.
- While the opposite is at least noticeable. If a user uses a custom image built on top of prior `21.2.13` version, the pods won't come up, he will see the error, check the Helm chart and disable the `securityContext`.

rail:
TBH I'm OK with both approaches, have no strong opinion. @prafull01 what do you think?

prafull01:
@tyranron has some good points about the custom image below `21.2.13` will have to disable the security context on their own based on errors. Those errors would be visible for the users and in my approach the user has to explicitly enable the security context else the pods would not have security context and user will not know about it without explicitly checking it.

I am good with the approach suggested by @tyranron. Could you please rebase your branch.

rail (approved):

PR #311, opened 2023-05-09
needs-build job should not fail when tagChange is not true

### LINKED ISSUE #310: helm-chart action fails on master branch
https://github.com/cockroachdb/helm-charts/actions?query=branch%3Amaster

Last passing run against `master` was July 2022: https://github.com/cockroachdb/helm-charts/actions/runs/2657195035

prafull01:
Fixes #310 

Added tagChange value explicitly false in job so that the job doesn't fail because of the return value from previous grep command.


celiala (approved):
LGTM - thank you for the fix!

PR #315, opened 2023-05-27
bazel: fix file rename

rail:


celiala (approved):
thanks for finding & fixing!

PR #332, opened 2023-07-18
Fix poddisruptionbudget api version switch

### LINKED ISSUE #331: no matches for kind PodDisruptionBudget in version policy/v1beta1
When trying to apply the Helm chart using server version v1.26.3 on AKS, the following error occurs:
no matches for kind PodDisruptionBudget in version policy/v1beta1 when

This appears to be broken: https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/templates/poddisruptionbudget.yaml#L2C1-L2C117

mmianl:
Closes #331

### DIFF
--- a/cockroachdb/templates/poddisruptionbudget.yaml
+++ b/cockroachdb/templates/poddisruptionbudget.yaml
@@ -1,5 +1,5 @@
 kind: PodDisruptionBudget
-{{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">= 1.21-0" .Capabilities.KubeVersion.Version) }}
+{{- if or (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version) }}
 apiVersion: policy/v1
 {{- else }}
 apiVersion: policy/v1beta1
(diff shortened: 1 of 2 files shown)


cockroachdb/templates/poddisruptionbudget.yaml
```diff
@@ -1,5 +1,5 @@
 kind: PodDisruptionBudget
-{{- if and (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">= 1.21-0" .Capabilities.KubeVersion.Version) }}
+{{- if semverCompare ">=1.21-0" .Capabilities.KubeVersion.GitVersion }}
```
prafull01, line 2: I believe this condition is failing because of the extra space in ">= 1.21-0". Once we remove the extra space the existing condition should pass with ">=1.21-0". Also, instead of `anding` these two conditions, we should use `or` operator to give a little extra cushion. So final condition should be:
```suggestion
{{- if or (.Capabilities.APIVersions.Has "policy/v1") (semverCompare ">=1.21-0" .Capabilities.KubeVersion.Version) }}
```

Can you please try with this condition and let me know if it passes on AKS.

prafull01 (changes requested):

cockroachdb/templates/poddisruptionbudget.yaml
mmianl: Yes, works that way.

prafull01 (approved):

mmianl:
@prafull01 thanks for your review and approval

PR #339, opened 2023-08-08
release: advance CRDB version to v22.2.13

THardy98:


celiala:
We only need to do this for 23.1

(We can close this PR -- no need to do for 22.2.13)

THardy98:
Not necessary for 22.2.x

PR #341, opened 2023-08-23
Add nameOverride and fullnameOverride parameters in values.yaml

### LINKED ISSUE #340: Can we please include `fullnameOverride` and `nameOverride` in the values.yaml?
Without `fullnameOverride` in your values.yaml in this repo, it sometimes can cause an error if a user tries to specify this and is using `helm template` ([Argo CD does this](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#declarative), for instance).

it's reference here:
https://github.com/cockroachdb/helm-charts/blob/9193c1fdd03c3dac3e71d2fe327ec3ae4e54a2a0/cockroachdb/templates/_helpers.tpl#L1-L42

Setting this in values.yaml would also make it more clear for users who are verifying which parameters are available.

Thank you for your time and for maintaining this chart! :)

prafull01:
Fixes #340 

Added the `nameOverride` and `fullnameOverride` parameters in values.yaml

### DIFF
--- a/build/templates/values.yaml
+++ b/build/templates/values.yaml
@@ -1,3 +1,9 @@
+# Overrides the chart name against the label "app.kubernetes.io/name: " placed on every resource this chart creates.
+nameOverride: ""
+
+# Override the resource names created by this chart which originally is generated using release and chart name.
+fullnameOverride: ""
+
 image:
   repository: cockroachdb/cockroach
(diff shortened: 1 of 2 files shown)


udnay (approved):
LGTM

rail (approved):
LGTM, :shipit:

PR #343, opened 2023-08-29
Add pod security policy

bkochendorfer:
This adds pod security policy and `seccompProfile`. I was unable to get the Helm chart to install locally to verify that these changes work when tls is enabled and the self signer is enabled, this seems to be the default configuration based on values. Whenever I attempt to do this I always get errors that the serviceaccount, roles, and rolebinding already exist due to how the annotations are being set manually. I'm not familiar enough with this chart to know if we should change those to include the release-name so that a subsequent run does not try to recreate them.

If I could get this to work locally I believe I will also need to resources in the jobs `job-certSelfSigner` and `job-cleaner` as we require setting resource limits/requests everywhere.

### DIFF
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -65,6 +65,17 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+        {{- if .Values.statefulset.securityContext.enabled }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
+            seccompProfile:
+              type: "RuntimeDefault"
+        {{- end }}
           volumeMounts:
             - name: certs
               mountPath: /cockroach-certs/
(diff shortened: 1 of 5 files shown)


udnay:
@himanshu-cockroach can you take a look at this?

cockroachdb/templates/statefulset.yaml
```diff
@@ -65,6 +65,17 @@ spec:
...
+        {{- if .Values.statefulset.securityContext.enabled }}
+          securityContext:
+            allowPrivilegeEscalation: false
+            capabilities:
+              drop:
+                - ALL
+            privileged: false
+            readOnlyRootFilesystem: true
+            seccompProfile:
+              type: "RuntimeDefault"
```
prafull01, line 77: This seccompProfile should be moved to the pod securityContext instead of keeping it in container securityContext which will be consistent to all other jobs securityContext

prafull01:
@bkochendorfer Please rebase your branch with latest change

cockroachdb/templates/statefulset.yaml
```diff
@@ -294,6 +303,8 @@ spec:
                 - ALL
             privileged: false
             readOnlyRootFilesystem: true
+            seccompProfile:
+              type: "RuntimeDefault"
```
prafull01, line 307: It is not needed here. Already added in pod securityContext

bkochendorfer: oops, sorry about that!

prafull01 (approved):

PR #363, opened 2024-01-09
Do not use docker images for build and release

rail:
Previously, we used a custom docker image with `helm` and `aws-cli` installed in order to generate Helm charts and to sync the generated files to S3. This docker image is very outdated and hosted on Docker Hub.

To simplify local development, remove docker dependency and keep the dependencies fresher, in this PR:

* Removed custom docker image and the wrapper script
* `helm` is installed run-time
* Use `aws-cli` installed system-wide
* Simplified docs

Epic: RE-539


celiala (approved):
LGTM! Nice cleanup :)

PR #364, opened 2024-01-09
Fix non zero exit code

rail:
`diff` returns non-zero exit code when there is a diff. We should not fail on this.


celiala (approved):
nice!

PR #369, opened 2024-01-09
Update dependencies

rail:
Some of the indirect dependencies contain CVE's, so it would be better to upgrade them.


rail:
duplicate of #370

PR #373, opened 2024-01-26
Add custom fields to self signer jobs

81887821:
Currently annotations, affinity, node selector and tolerations are only applied to self signer init job. It would be better if those are also applied to rotate cronjobs and clean-up job.
My use case is to disable Istio sidecar injection to the clean-up job.

### DIFF
--- a/build/templates/values.yaml
+++ b/build/templates/values.yaml
@@ -529,6 +529,9 @@ tls:
       nodeCertExpiryWindow: 168h
 
   selfSigner:
+    # Additional labels to apply to the Pod of this Job.
+    labels: {}
+
     # Additional annotations to apply to the Pod of this Job.
     annotations: {}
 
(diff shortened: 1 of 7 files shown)


umi0410:
It will be great supporting both additional labels and annotations.
For your reference, the current related code is here: https://github.com/cockroachdb/helm-charts/blob/0cba0b68def62547888f52d6183a736af4c06292/cockroachdb/templates/job-cleaner.yaml#L20-L26

81887821:
Added support for labels

udnay:
@prafull01 can you take a look at this please?

@81887821 can you please rebase with master?

prafull01 (approved):

prafull01:
LGTM, Approved. @81887821 Please rebase again, so that I can merge this change.

PR #378, opened 2024-02-26
Update terminationGracePeriodSeconds to 300 seconds

prafull01:
We've extended the terminationGracePeriodSeconds to 300 seconds for the CockroachDB StatefulSet, allowing up to 300 seconds for termination of the pod. Consequently, we need to adjust the Helm upgrade timeout accordingly for a standard 3-node CockroachDB cluster.


rail (approved):
LGTM, :shipit:

PR #390, opened 2024-05-23
Automerge version bumps

rail:
This change marks PRs created by `cockroach-teamcity` ready to be merged after all checks are passed.

### DIFF
--- a/.github/workflows/automerge.yaml
+++ b/.github/workflows/automerge.yaml
@@ -1,0 +1,63 @@
+# This Workflow automatically merges pull requests, submitted by certain users
+# if they pass certain criteria.
+name: Automerge version bumps
+
+on:
+  pull_request:
+    branches: [ master ]
+    paths:
+    # Trigger the workflow only for certain files. First we need to exclude all
+    # files, then add the files we care about.
+    - "!**"
+    - "cockroachdb/Chart.yaml"
+    - "cockroachdb/README.md"
+    - "cockroachdb/values.yaml"
+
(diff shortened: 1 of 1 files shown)


.github/workflows/automerge.yaml
```diff
@@ -0,0 +1,63 @@
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          # Fetch all branches so it is possible to checkout the default branch files.
+          fetch-depth: 0
+      # The next steps tries to reproduce the steps taken to generate the
+      # files. We restore the changed files to their original state on the
+      # default branch, run the commands to regenerate the change, and verify
+      # the result matches the PR contents (git diff does not show any changes).
+      - name: Regenerate the patch
```
celiala, line 33: nit: maybe name this "confirm expected patch" or something?

(then under "verify nothing changed" move the build steps in there, i.e.)

>  bazel build //build
> $(bazel info bazel-bin)/build/build_/build bump "$appVersion"

celiala (approved):
woo - excited for this!

had one suggestion (inline) - but not blocking (i.e. PR also okay as-is)

.github/workflows/automerge.yaml
rail: I'll rename the step. 👍 
The build steps generate the patch, so they kind of belong to this step IMO.

celiala: ah, gotcha. cool, works for me :) 👍🏼

Issue #407, opened 2024-08-22
Let user to customize terminationGracePeriodSeconds


lin-crl:
Currently it's hardcoded at 300s - 5 minutes in [Statefulset](https://github.com/cockroachdb/helm-charts/blob/7d9b09550157f5d7d5987b4a0044d2639ed8abad/cockroachdb/templates/statefulset.yaml#L151)
It's too long for Dev and too short for Production 

```
terminationGracePeriodSeconds: 300
```

I'd like to request it to make it customizable using values.yaml.
Thanks!

PR #416, opened 2024-09-18
Allow terminationGracePeriodSeconds to be configured via helm values

prafull01:
Fixes: #407

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -49,7 +49,7 @@ spec:
     {{- end }}
     {{- end }}
       restartPolicy: OnFailure
-      terminationGracePeriodSeconds: 300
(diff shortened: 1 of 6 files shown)


rail (approved):

cockroachdb/templates/job.init.yaml
```diff
@@ -49,7 +49,7 @@ spec:
     {{- end }}
     {{- end }}
       restartPolicy: OnFailure
-      terminationGracePeriodSeconds: 300
+      terminationGracePeriodSeconds: {{ .Values.init.terminationGracePeriodSeconds }}
```
udnay, line 52: If we are adding this, I would add it to the readme

prafull01: Done

PR #419, opened 2024-09-23
Allow changing of grpc and http ports from single placeholder

prafull01:


cockroachdb/templates/statefulset.yaml
```diff
@@ -202,8 +202,8 @@ spec:
             {{- with .Values.conf.attrs }}
               --attrs={{ join `:` . }}
             {{- end }}
-              --http-port={{ index .Values.conf `http-port` | int64 }}
-              --port={{ .Values.conf.port | int64 }}
+              --http-port={{ index .Values.service.ports.http.port | int64 }}
```
udnay, line 205: If someone has `conf.port` and/or `conf.http-port` set, this change will break the chart, right? Could we set them to `""` in `values.yaml` and only use them if they are set somewhere explicitly? Maybe something like.

```
{{- if .Values.conf.port }}
  --port={{ .Values.conf.port | int64 }} #WARNING this parameter is deprecated and will be removed in a future version. Use `.Values.service.ports.grpc.internal` instead
{{ - else }}
--port={{ .Values.service.ports.grpc.internal.port | int64 }}
{{ - end }}
```
wdyt?

prafull01: Yes, that would generally be true in most cases. However, for the Cockroach cluster to function correctly here, `.Values.conf.port` must match `.Values.service.ports.grpc.internal.port`, and similarly, `.Values.conf.http-port` must match `.Values.service.ports.http.port`.

For any cluster in a ready state, both sets of values should already be the same. But for added safety, I'm okay with keeping the old placeholders as a precaution.

prafull01: Added the changes.

udnay (changes requested):
Please update the README as well https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/README.md?plain=1#L311-L312

build/templates/README.md
```diff
@@ -307,8 +307,6 @@ For details see the [`values.yaml`](values.yaml) file.
 | `conf.locality`                                           | Locality attribute for this deployment                          | `""`                                                  |
 | `conf.single-node`                                        | Disable CockroachDB clustering (standalone mode)                | `no`                                                  |
 | `conf.sql-audit-dir`                                      | Directory for SQL audit log                                     | `""`                                                  |
-| `conf.port`                                               | CockroachDB primary serving port in Pods                        | `26257`                                               |
-| `conf.http-port`                                          | CockroachDB HTTP port in Pods                                   | `8080`                                                |
 | `conf.path`                                               | CockroachDB data directory mount path                           | `cockroach-data`                                      |
```
udnay, line 310:
```suggestion
| `conf.port`                                               | WARNING this parameter is deprecated and will be removed in a future version. Use `.service.ports.grpc.internal.port` instead                       | ""                                               |
| `conf.http-port`                                          | WARNING this parameter is deprecated and will be removed in a future version. Use `.service.ports.http.port` instead                                   | ""                                                |
```

cockroachdb/README.md
```diff
@@ -308,8 +308,6 @@ For details see the [`values.yaml`](values.yaml) file.
 | `conf.locality`                                           | Locality attribute for this deployment                          | `""`                                                  |
 | `conf.single-node`                                        | Disable CockroachDB clustering (standalone mode)                | `no`                                                  |
 | `conf.sql-audit-dir`                                      | Directory for SQL audit log                                     | `""`                                                  |
-| `conf.port`                                               | CockroachDB primary serving port in Pods                        | `26257`                                               |
-| `conf.http-port`                                          | CockroachDB HTTP port in Pods                                   | `8080`                                                |
 | `conf.path`                                               | CockroachDB data directory mount path                           | `cockroach-data`                                      |
```
udnay, line 311:
```suggestion
| `conf.port`                                               | WARNING this parameter is deprecated and will be removed in a future version. Use `.service.ports.grpc.internal.port` instead                       | ""                                               |
| `conf.http-port`                                          | WARNING this parameter is deprecated and will be removed in a future version. Use `.service.ports.http.port` instead                                   | ""                                                |
```

udnay (approved):
Suggested some changes to add the deprecation warning to the readme as well. Other than that LGTM

udnay (approved):
LGTM

PR #426, opened 2024-11-06
Swap kind for k3d

udnay:
Use k3d so that we are standard for our local k8s dev environment across projects


udnay:
@prafull01 can you take a look?

rail (approved):
LGTM

PR #436, opened 2024-12-11
[CRDB-44997] feature: support WAL failover

pritesh-lahoti:
Adds support for WAL failover via:
- among multiple stores
- single store side disk (as a PVC)

### DIFF
--- a/cockroachdb/templates/statefulset.yaml
+++ b/cockroachdb/templates/statefulset.yaml
@@ -1,4 +1,5 @@
 {{ template "cockroachdb.conf.log.validation" . }}
+{{ template "cockroachdb.conf.store.validation" . }}
 kind: StatefulSet
 apiVersion: {{ template "cockroachdb.statefulset.apiVersion" . }}
 metadata:
@@ -235,7 +236,14 @@ spec:
               --sql-audit-dir={{ . }}
             {{- end }}
             {{- if .Values.conf.store.enabled }}
-              --store={{ template "cockroachdb.conf.store" . }}
+              {{- range $idx := until (int .Values.conf.store.count) }}
+              {{- $_ := set $ "Args" (dict "idx" $idx) }}
+              --store={{ include "cockroachdb.conf.store" $ }}
+              {{- end }}
+            {{- end }}
+            {{- with index .Values.conf `wal-failover` `value` }}
+              {{- template "cockroachdb.conf.wal-failover.validation" $ }}
+              --wal-failover={{ . }}
             {{- end }}
             {{- if .Values.conf.log.enabled }}
               --log-config-file=/cockroach/log-config/log-config.yaml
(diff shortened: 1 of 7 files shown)


cockroachdb/values.yaml
```diff
@@ -176,19 +176,60 @@ conf:
...
     # Arbitrary strings, separated by colons, specifying disk type or capability
     attrs:
 
+  # CockroachDB's WAL failover configuration:
+  # https://www.cockroachlabs.com/docs/stable/cockroach-start#write-ahead-log-wal-failover
+  # Uses `--wal-failover` flag
+  wal-failover:
+    # The value to be passed to the `--wal-failover` flag.
+    # Possible configurations:
+    # 1. Default: If empty, `--wal-failover` is not passed to cockroach start
```
BabuSrithar, line 205: Nit: we can have values in the beginning to make it more readable

Possible configurations : 
1. `` : When empty, the configuration is ignored and  `--wal-failover` is not passed to cockroach start, this is also the default behavior. 
2. `disabled` : Disables WAL failover. 
3. `among-stores` : Enables WAL failover  among multiple stores. Setting `conf.store.count` greater than 1 is a dependency for this to to take effect. 
4. `path=<path-to-side-disk>`: Enables WAL failover on a side disk. A persistent volume should be mounted at the specified path (e.g. `path=/cockroach/cockroach-failover`).

pritesh-lahoti: Addressed.

cockroachdb/templates/statefulset.yaml
```diff
@@ -344,16 +365,42 @@ spec:
...
+        {{- end }}
+      {{- else }}
+        - name: datadir-{{ add1 $i }}
+        {{- if $.Values.storage.persistentVolume.enabled }}
+          persistentVolumeClaim:
+            claimName: datadir-{{ add1 $i }}
+        {{- else if $.Values.storage.hostPath }}
           hostPath:
-            path: {{ .Values.storage.hostPath | quote }}
+          path: {{ $.Values.storage.hostPath | quote }}
```
prafull01, line 387: hostPath is a directory on a host (node on which cockroachdb pod is running) where a particular volume is mounted. This means the cockroachdb pod writing the data is going on the host at that particular path.

In above example, all the datadir is going on the same hostPath on the same host. I think this will cause issues when multi store is used with hostPath

prafull01: Also `hostPath` indentation on line 387 is also incorrect

prafull01 (commented):
We should add upgrade documentation when number of datastore changes as it is not traditional upgrade for user.

We should also add an e2e test which covers the migration scenario with upgrade. If we decide to do it later, please add a ticket for the same

cockroachdb/templates/statefulset.yaml
pritesh-lahoti: Fixed. Thank you!

prafull01 (approved):

pritesh-lahoti:
> We should add upgrade documentation when number of datastore changes as it is not traditional upgrade for user.
> 
> We should also add an e2e test which covers the migration scenario with upgrade. If we decide to do it later, please add a ticket for the same

Will track the documentation separately. Updated the PR with an e2e test.

prafull01 (approved):

PR #438, opened 2024-12-18
[CRDB-44997] docs: update docs to upgrade chart involving new PVCs

pritesh-lahoti:
The upgrade process for a chart involving new PVCs is manual and needs to be well documented. We will be working on automating it soon.

This PR also fixes other minor docs and updates the release.


build/templates/README.md
```diff
@@ -293,145 +304,148 @@ Note that there are sometimes backward-incompatible changes to SQL features betw
...
+| `networkPolicy.ingress.grpc`                              | Whitelist resources to access gRPC port of CockroachDB's Pods                                                                                                                                                                                                                                                                            | `[]`                                                   |
+| `networkPolicy.ingress.http`                              | Whitelist resources to access gRPC port of CockroachDB's Pods                                                                                                                                                                                                                                                                            | `[]`                                                   |
```
prafull01, line 440: Nit: This is one line change for `conf.wal-failover` so if you carefully change then the diff should only show one line as follows:

This would be helpful if we look at this PR later.

prafull01 (approved):

build/templates/README.md
pritesh-lahoti: I actually added `conf.log` and `conf.store.count` too.

Also, `conf.locality` was incorrectly placed, so had to move it correctly. I had carefully made these changes and the resulting diff is what you see.

I realize that we are not strictly following the lexicographical order (e.g. `conf.port`, `conf.http-port`), so I'll undo this change and see what diff it generates.

PR #439, opened 2024-12-18
test: fix flaky test due to timeout

pritesh-lahoti:
The e2e test times out probably due to a lot of PVs getting provisioned and running out of space.

This PR restricts the PV sizes across all tests to 1Gi (as opposed to the default 100Gi). It also now uses the ubuntu-latest-4-core runner (with higher storage) for the e2e tests.

### DIFF
--- a/tests/e2e/install/cockroachdb_helm_e2e_test.go
+++ b/tests/e2e/install/cockroachdb_helm_e2e_test.go
@@ -59,31 +59,27 @@ func TestCockroachDbHelmInstall(t *testing.T) {
 	// Setup the args. For this test, we will set the following input values:
 	options := &helm.Options{
 		KubectlOptions: k8s.NewKubectlOptions("", "", namespaceName),
-		SetValues: map[string]string{
+		SetValues: patchHelmValues(map[string]string{
 			"conf.cluster-name":                        "test",
 			"init.provisioning.enabled":                "true",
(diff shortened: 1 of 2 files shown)


tests/e2e/install/cockroachdb_helm_e2e_test.go
```diff
@@ -639,3 +600,38 @@ func testWALFailoverExistingCluster(t *testing.T, additionalValues map[string]st
...
 }
+
+func getHelmValues(inputValues map[string]string) map[string]string {
+	defaultValues := map[string]string{
+		// Override the persistent storage size to 1Gi so that we do not run out of space.
+		"storage.persistentVolume.size": "1Gi",
+	}
+
+	for k, v := range defaultValues {
+		inputValues[k] = v
```
rail, line 611: I would expect this function to merge the default values with the passed values, using the passed values as overrides. This way you would be able to set the storage size to something other than 1Gi.
maybe change the function to something like:
```go
for k, v:= range inputValues {
	defaultValues[k] = v
}
return defaultValues
```

pritesh-lahoti: I had actually intended to override the input values. I could maybe update the variable/function names to make this clear.
I don't see a reason why a test would need >1Gi persistent storage. If it comes to that, will change it later.

rail (approved):
LGTM, :shipit:

PR #442, opened 2025-01-09
fix default resource names.

jmcarp:
The default names of service accounts, secrets, and configmaps currently don't match the defaults in the crdbcluster template. This patch updates the crdbcluster defaults such that creating a new cluster works without overrides.


pritesh-lahoti (approved):
Thank you!

PR #443, opened 2025-01-10
CRDB-45670: helm: automate the statefulset update involving new PVCs

pritesh-lahoti:
In scenarios where a new PVC is getting added to a statefulset as part of Helm upgrade, we need to perform the upgrade as per the below steps:
For every pod:

1. delete the statefulset with --cascade=orphan
2. delete the pod
3. perform helm upgrade

This is required as we cannot attach a volume to an existing pod.

It is not reasonable to expect customers to manually perform these steps as part of their Helm upgrade, so we would like to offer a script which could automate the same and reduce friction for the customers.

### DIFF
--- a/scripts/upgrade_with_new_pvc.sh
+++ b/scripts/upgrade_with_new_pvc.sh
@@ -1,0 +1,67 @@
+#!/bin/bash
+
+Help()
+{
+   # Display Help
+   echo "This script performs Helm upgrade involving new PVCs. Kindly run it from the root of the repository."
+   echo
+   echo "usage: ./scripts/upgrade_with_new_pvc.sh <release_name> <chart> <chart_version> <values_file> <namespace> <sts_name> <num_replicas> [kubeconfig]"
+   echo
+   echo "options:"
+   echo "release_name: Helm release name, e.g. my-release"
+   echo "chart: Helm chart to use (either referenced locally, or to the Helm repository), e.g. cockroachdb/cockroachdb"
+   echo "chart_version: Helm chart version to upgrade to, e.g. 15.0.0"
+   echo "values_file: Path to the values file, e.g. ./cockroachdb/values.yaml"
+   echo "namespace: Kubernetes namespace, e.g. default"
+   echo "sts_name: Statefulset name (can be obtained through \"kubectl get sts\"), e.g. my-release-cockroachdb"
+   echo "num_replicas: Number of replicas in the statefulset, e.g. 3"
+   echo "kubeconfig (optional): Path to the kubeconfig file. Default is $HOME/.kube/config."
+   echo
+   echo "example: ./scripts/upgrade_with_new_pvc.sh my-release cockroachdb/cockroachdb 15.0.0 ./cockroachdb/values.yaml default my-release-cockroachdb 3"
+   echo
+}
+
+while getopts ":h" option; do
+   case $option in
+      h) # display Help
+         Help
+         exit;;
+     \?) # incorrect option
+         echo "Error: Invalid option"
+         exit;;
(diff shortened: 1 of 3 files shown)


scripts/upgrade_with_new_pvc.sh
```diff
@@ -0,0 +1,60 @@
...
+   echo
+   echo "options:"
+   echo "release_name: Helm release name, e.g. my-release"
+   echo "chart_version: Helm chart version to upgrade to, e.g. 15.0.0"
+   echo "namespace: Kubernetes namespace, e.g. default"
+   echo "sts_name: Statefulset name, e.g. my-release-cockroachdb"
+   echo "num_replicas: Number of replicas in the statefulset, e.g. 3"
+   echo "kubeconfig (optional): Path to the kubeconfig file. Default is $HOME/.kube/config."
+   echo
+   echo "example: ./scripts/upgrade_with_new_pvc.sh my-release 15.0.0 default my-release-cockroachdb 3"
```
prafull01, line 18: We should also take an input for values.yaml file. User could have custom values.yaml file they created for cockroachdb. User could be passing this custom values.yaml file using `-f` option in helm upgrade command.

scripts/upgrade_with_new_pvc.sh
```diff
@@ -0,0 +1,60 @@
...
+
+  echo "$((i+1)). Deleting replica"
+  kubectl --kubeconfig=$kubeconfig -n $namespace delete pod $sts_name-$i --wait=true
+
+  echo "$((i+1)). Upgrading Helm"
+  # The "--wait" flag ensures the deleted pod replica and STS are up and running.
+  # However, at times, the STS fails to understand that all replicas are running and the upgrade is stuck.
+  # The "--timeout 1m" helps with short-circuiting the upgrade process. Even if the upgrade does time out, it is
+  # harmless and the last upgrade process will be successful once all the pods replicas have been updated.
+  helm upgrade $release_name ./cockroachdb --kubeconfig=$kubeconfig --namespace $namespace --version $chart_version --wait --timeout 1m --debug
```
prafull01, line 59: If one replica is not upgraded and not joined to the cockroachdb cluster properly and we move to the next one. Wouldn't it affect the quoram of the cluster if there are 3 nodes and 2 nodes are not accessible at the moment?

We should identify that the replica we have updated is joined to the cluster before moving to the next one.

scripts/upgrade_with_new_pvc.sh
pritesh-lahoti: At L59, I'm currently using `./cockroachdb` as the chart location to upgrade to. I am guessing this would default to using the values file at `./cockroachdb/values.yaml`, right?
Once we do provide a flag to provide the values file (`-f`), I am thinking if it's okay to keep the chart path as `./cockroachdb`.

scripts/upgrade_with_new_pvc.sh
pritesh-lahoti: Depending on the customer's configuration, they may need to perform specific sets of validations.
Have added a manual prompt to defer this onto the customer.

scripts/upgrade_with_new_pvc.sh
prafull01: Keeping the path as `./cockroachdb` could be an issue. Customer is not necessarily using the github repo to install the charts. Most of the customers are using the helm repo to install. In such case, this could be an issue.

pritesh-lahoti: Added arguments corresponding to the chart and values file both.

prafull01 (approved):

PR #446, opened 2025-01-15
cockroachdb: add crdbcluster options.

jmcarp:
Add options to the crdbcluster template that we set in cloud, along with sensible defaults.

### DIFF
--- a/cockroachdb/values.yaml
+++ b/cockroachdb/values.yaml
@@ -822,6 +822,25 @@ operator:
   podLabels:
     app.kubernetes.io/component: cockroachdb
 
+  # Flags passed to the cockroachdb container.
+  flags:
+    # Disable backup/restore to local disk by default.
+    --external-io-dir: disabled
+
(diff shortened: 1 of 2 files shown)


cockroachdb/values.yaml
```diff
@@ -822,6 +822,25 @@ operator:
...
+  # unavailability during restarts.
+  rollingRestartDelay: 3m0s
+
+  # Topology spread constraints set on cockroachdb pods. Spread cockroachdb
+  # pods across zones by default.
+  topologySpreadConstraints:
+    - maxSkew: 1
+      topologyKey: topology.kubernetes.io/zone
+      whenUnsatisfiable: DoNotSchedule
+
```
pritesh-lahoti, line 843: You'll need to update `build/templates/values.yaml` and then run `go run build/build.go generate` to generate the changes to this file.

pritesh-lahoti (approved):
Pre-approving

PR #447, opened 2025-01-17
Check in initial e2e tests for cloud operator.

jmcarp:
Add a test to deploy the operator, then a single-node crdb cluster, and test basic functionality.

### DIFF
--- a/tests/testutil/require.go
+++ b/tests/testutil/require.go
@@ -35,13 +35,12 @@ type CockroachCluster struct {
 	ClientSecret, NodeSecret   string
 	CaSecret                   string
 	IsCaUserProvided           bool
+	DesiredNodes               int
 }
 
 // RequireClusterToBeReadyEventuallyTimeout waits for all the CRDB pods to come into running state.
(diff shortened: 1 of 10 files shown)


tests/testutil/require.go
```diff
@@ -161,15 +207,75 @@ func RequireDatabaseToFunction(t *testing.T, crdbCluster CockroachCluster, dbNam
...
+	db := getDBConn(t, crdbCluster, "system", "")
 
 	if rotate {
 		t.Log("Verifying the existing data in the database after certificate rotation")
 	}
 
 	// Create database only if we are testing crdb install
 	if !rotate {
-		if _, err := db.Exec("CREATE DATABASE test_db"); err != nil {
+		if _, err := db.Exec("CREATE DATABASE IF NOT EXISTS testdb"); err != nil {
```
pritesh-lahoti, line 218:
```suggestion
		if _, err := db.Exec("CREATE DATABASE IF NOT EXISTS test_db"); err != nil {
```
Looks like this is causing the test to fail on CI.

pritesh-lahoti (approved):
Pre-approving

tests/testutil/require.go
jmcarp: Thanks, fixed!

Issue #402, opened 2024-07-18
CockroachDB doesnt start and init doesnt seem to launch


DelaunayAntoine:
Hello everyone, 

I would like to deploy cockroachDB using helm but the problem is that the cluster can't start and I get this error that keeps appearing: 
Error `I240718 13:09:06.023104 191 server/init.go:405 ⋮ [T1,Vsystem,n?] 37  ‹cockroachdb-1.cockroachdb.cockroachdb.svc.cluster.local:26257› is itself waiting for init, will retry
`

Can you help me by giving me some hints on how to fix the problem? 

Here's the entire log file and the values.yaml file
[db.txt](https://github.com/user-attachments/files/16284048/db.txt)
[values-cockroach.txt](https://github.com/user-attachments/files/16284087/values-cockroach.txt)

I'm using cockroach version 24.1.1 
The chart 13.0.1

**What do you expect to see ?**

The cockroach cluster launching just fine

**What happened**

Error `I240718 13:09:06.023104 191 server/init.go:405 ⋮ [T1,Vsystem,n?] 37  ‹cockroachdb-1.cockroachdb.cockroachdb.svc.cluster.local:26257› is itself waiting for init, will retry
`

lknite:
I'm seeing this as well.  Did you figure it out?

apavarnitsyn:
I've got the same problem with the latest 14.0.3 chart.
I suppose that the reason is in helm hooks annotations of init job template.
https://github.com/cockroachdb/helm-charts/blob/master/cockroachdb/templates/job.init.yaml#L22
Post-install hook can't be triggered because the stateful set is not ready.
As a workaround you may deploy the init job manifest from template manually.

udnay:
Are either of you able to share your values file? A redacted version is likely fine, just to see what overrides you have set. I have. not ben able to reproduce this with the default values.

lknite:
@udnay , here ya go:
```
$ cat Chart.yaml 
apiVersion: v2
name: jellyfin
description: A Helm chart for Kubernetes

# A chart can be either an 'application' or a 'library' chart.
#
# Application charts are a collection of templates that can be packaged into versioned archives
# to be deployed.
#
# Library charts provide useful utilities or functions for the chart developer. They're included as
# a dependency of application charts to inject those utilities and functions into the rendering
# pipeline. Library charts do not define any templates and therefore cannot be deployed.
type: application

# This is the chart version. This version number should be incremented each time you make changes
[...]

Issue #287, opened 2022-12-14
Helm chart will never succeed when `helm install` is run with `--atomic` flag set


worldofgeese:
When `helm install --set storage.persistentVolume.size="25Gi" my-cockroachdb cockroachdb/cockroachdb --version 10.0.0 --atomic` is run, the Helm chart install will never succeed and hang until timeout is exceeded, at which point the chart uninstalls. It appears at least the `cockroachdb-init` Job is never run (but I believe this extends to the 2 other cert Jobs) with `--atomic` but does run without. This pod is critical to setting the cluster up for the first time.

jonasbadstuebner:
> The --wait flag will be set automatically if --atomic is used

ref: https://helm.sh/docs/helm/helm_install/#options

You probably want to have a look at this issue and my comment there: https://github.com/cockroachdb/helm-charts/issues/402#issuecomment-2605920759
If this solves your problem, feel free to close this issue, as it would be a duplicate.

Issue #69, opened 2021-04-13
Job is not run on first install


PPACI:
Hi cockroachDB.

I try to install a new cluster.

My pod are all stuck waiting for an init job.

I think it come from the job being a post install helm hook.
As the pod are not ready (waiting for the job), the install will never be done and thus the job never run.

https://github.com/cockroachdb/helm-charts/blob/a20ee1eb40168fd5125f691fff9f762e43a3cf8e/cockroachdb/templates/job.init.yaml#L19

How could I install a brand new cluster?

DuskEagle:
These hooks were added to solve #68. Previously, you had to manually remove the init job after it ran, or you would encounter a problem when attempting to update your CockroachDB version via Helm.

One workaround to this issue is to not use the `--wait` flag. How critical is this flag to your use of Helm?

The only other thing I could see would be making the hooks optional, so you could choose between having to delete the init job after installing, or choose whether to not use the `--wait` flag.

PPACI:
I confirm that not using `--wait` make it work.
For reference. Terraform add a `wait=true` by default on each Helm release. This was my setup.

In my case, it's not critical and I can live without `--wait`.

Maybe something like

```
init:
  forceJob: false
```
In values.

Setting to false (the default) would have the current behaviour. But setting to true would force create a job that you would have to manually cleanup.

I have a couple of other stuff to do with this chart to support some of our "enterprise" use case.
If everything goes right, I can open a couple of PR, including this switch.

Would that work for you?

dpkirchner:
Using `wait = false` and `wait_for_jobs = true` in Terraform seems to work for me, although it all happens fast enough that I wouldn't be able to absolutely guarantee that the init `Job` was indeed done before Terraform finished. I've eyeballed the events log and it *seems* to be doing what I expected -- only after cockroachdb-0 (first STS node) was live did the job complete and did Helm (or Terraform?) mark the release as deployed.

`wait_for_jobs` is a relatively new flag, added shortly after this issue was opened.

Issue #410, opened 2024-08-26
Clean deployment of helm chart doesn't result in working cockroachdb instance.


lknite:
I'm wondering if the helm chart is failing for everyone?

I haven't set anything in my values.yaml so this is a clean installation of the cockroachdb helm chart into a new namespace.

I'm seeing this in the logs:
```
[keycloak-cockroachdb-1 db] I240826 18:06:22.620811 146 server/init.go:451 ⋮ [T1,Vsystem,n?] 4095  ‹keycloak-cockroachdb-2.keycloak-cockroachdb.keycloak.svc.cluster.local:26257› is itself waiting for init, will retry
[keycloak-cockroachdb-0 db] I240826 18:06:23.505955 147 server/init.go:451 ⋮ [T1,Vsystem,n?] 170  ‹keycloak-cockroachdb-2.keycloak-cockroachdb.keycloak.svc.cluster.local:26257› is itself waiting for init, will retry
[keycloak-cockroachdb-2 db] I240826 18:06:23.540993 158 server/init.go:451 ⋮ [T1,Vsystem,n?] 4094  ‹keycloak-cockroachdb-0.keycloak-cockroachdb.keycloak.svc.cluster.local:26257› is itself waiting for init, will retry
[keycloak-cockroachdb-1 db] I240826 18:06:23.615954 146 server/init.go:451 ⋮ [T1,Vsystem,n?] 4096  ‹keycloak-cockroachdb-0.keycloak-cockroachdb.keycloak.svc.cluster.local:26257› is itself waiting for init, will retry
```

what's running:
```
$ k get all
NAME                                         READY   STATUS             RESTARTS         AGE
pod/keycloak-0                               0/1     CrashLoopBackOff   16 (3m33s ago)   74m
pod/keycloak-cockroachdb-0                   0/1     Running            0                9m11s
pod/keycloak-cockroachdb-1                   0/1     Running            0                74m
pod/keycloak-cockroachdb-2                   0/1     Running            0                74m
pod/keycloak-cockroachdb-self-signer-w697r   0/1     Completed          0                75m

NAME                                  TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)              AGE
service/keycloak                      ClusterIP   10.96.22.110    <none>        80/TCP               74m
service/keycloak-cockroachdb          ClusterIP   None            <none>        26257/TCP,8080/TCP   74m
[...]

PR #450, opened 2025-01-21
fix: Make init-job run as hook only on install

jonasbadstuebner:
There are multiple issues stating the init-job is not run as expected

This is because there is a deadlock with the CRDB-StatfulSet requiring the init-job to run, which is only ran by Helm when the StatefulSet is considered ready

The optional use of the --wait of `helm install` is causing differing observations

This PR attempts to fix the problem by using the Job as plain Job instead of hook when the Chart is initially installed

### DIFF
--- a/cockroachdb/templates/job.init.yaml
+++ b/cockroachdb/templates/job.init.yaml
@@ -19,12 +19,17 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   annotations:
-    helm.sh/hook: post-install,post-upgrade
+    {{- if not .Release.IsInstall }}
+    helm.sh/hook: post-upgrade
     helm.sh/hook-delete-policy: before-hook-creation
+    {{- end }}
     {{- with .Values.init.jobAnnotations }}
     {{- toYaml . | nindent 4 }}
     {{- end }}
 spec:
+  {{- if .Release.IsInstall }}
+  ttlSecondsAfterFinished: 0
+  {{- end }}
   template:
     metadata:
       labels:


jonasbadstuebner:
⚠ Not tested yet ⚠ 
The rendering looks good - I am just not sure if this behaves as I think it does.
If someone else wants to test it, feel free.

Expected behavior:
On install (`--wait` given or not does not matter):
```md
1. The CRDB-StatefulSet and the init-Job are deployed almost at the same time
2. The init-job initializes the DB - once it is ready for initialization - and finishes
3. The `helm install` command succeeds
4. The Job is eventually auto-deleted by Kubernetes due to the `ttlSecondsAfterFinished` being 0
   (this unblocks the `helm upgrade`)
```

On Upgrade:
```md
(0. The init-Job is removed, if present from a previous `helm upgrade` - should not be
    there anymore from the `helm install` step)
1. The CRDB-StatefulSet is upgraded
2. The init-Job is run after the StatefulSet is upgraded and ready
3. The `helm upgrade` command succeeds
```

jonasbadstuebner:
**The open issues related to the init-job deadlock:**
For sure:
https://github.com/cockroachdb/helm-charts/issues/69
https://github.com/cockroachdb/helm-charts/issues/287
https://github.com/cockroachdb/helm-charts/issues/402

Probably:
https://github.com/cockroachdb/helm-charts/issues/345 (same error message)

**The closed issues related to the init-job deadlock:**
https://github.com/cockroachdb/helm-charts/issues/195
https://github.com/cockroachdb/helm-charts/issues/234
https://github.com/cockroachdb/helm-charts/issues/389 (I'd argue this is the same topic, even though the person put the blame on terraform)
https://github.com/cockroachdb/helm-charts/issues/400
https://github.com/cockroachdb/helm-charts/issues/410

This just shows that the topic is highly confusing, as people expect a simple `helm install` to work. I really hope my solution works.
Otherwise, I would love to see a big note in the README at least, to not use `--wait` on `helm install`.

jonasbadstuebner:
**I just had big success with flux and the init job.**

The changes I propose in this PR seem to do their job (pun intended...?).

The cluster was initialized without any manual action.
This is my install and upgrade section for the Fluxv2 HelmRelease:
```
apiVersion: helm.toolkit.fluxcd.io/v2
kind: HelmRelease
[...]
spec:
  install:
    remediation:
      retries: -1
  upgrade:
    remediation:
      retries: -1
```
Nothing special required here, as you can see. Flux enables `--wait` and `--wait-for-jobs` per default, so these flags should work on the normal `helm install` command after my changes too.

The init-job was applied in parallel to the StatefulSet, ran in the loop until the CRDB-pods accepted the init-command, ran the init, printed a success message and stayed as "completed" (which is expected and totally fine) before I got the `Helm install succeeded` message.

And when the upgrade ran, it removed the previously installed job - because of the `helm.sh/hook-delete-policy: before-hook-creation` annotation - and it ran the job after the StatefulSet was fully restarted.

Side-Note: It is probably debatable, if the init-job is even needed at all on upgrade...because what does it do there?

I mark this PR as ready and would be happy to address any questions or notes regarding my code changes.

jonasbadstuebner:
I'm unsure what a good title for this PR would be, so if you have one, let me know please

jonasbadstuebner:
closes #69 
closes #287 
closes #402 
closes #345 (I think - as I said, it is the same error message and the behavior is the same as when the init-job is not run because the StatefulSet did not become ready - also the person used `--wait=true` in [their install command](https://github.com/cockroachdb/helm-charts/issues/345#issuecomment-1709903478))

cockroachdb/templates/job.init.yaml
```diff
@@ -97,57 +99,32 @@ spec:
...
           # Command completes either when cluster initialization succeeds,
           # or when cluster has been initialized already.
           command:
           - /bin/bash
           - -c
-          - >-
+          - |
             {{- if $isClusterInitEnabled }}
-              initCluster() {
-                while true; do
```
prafull01, line 114: Removing this loop can cause issues, what if the init was not successful in first attempt. Job will fail and cockroachdb will not come up.

cockroachdb/templates/job.init.yaml
```diff
@@ -19,8 +19,10 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
   annotations:
-    helm.sh/hook: post-install,post-upgrade
-    helm.sh/hook-delete-policy: before-hook-creation
+    {{- if not .Release.IsInstall }}
+    helm.sh/hook: post-upgrade
+    helm.sh/hook-delete-policy: before-hook-creation,hook-succeeded
```
prafull01, line 24: The main issue with this is the job will be left in the cluster and when you do an upgrade, the new image will be patched to this job which is not allowed as job is an immutable resource in kubernetes.

cockroachdb/templates/job.init.yaml
jonasbadstuebner: `cockroach init` already runs in a loop. I think this was changed since the Job was created and the Job was never updated..

cockroachdb/templates/job.init.yaml
jonasbadstuebner: Since the `hook-delete-policy` still applies on Upgrade, it removes the old job before creating the one for the upgrade.
Feel free to test it out.

PR #459, opened 2025-02-14
scripts: add migration script from public operator to cloud operator.

jmcarp:
Check in a reference implementation for migrating from statesets managed by the public operator to the cloud operator. Note that this process involves some manual steps, and we may want to automate and test it further.


prafull01 (commented):
Changes are working as expected with some modifications. I have left those as comments. After this many comments, I'm half expecting you to blacklist me 😆

migrate/main.go
```diff
@@ -0,0 +1,303 @@
...
+			ExternalCertificates: &v1alpha1.ExternalCertificates{
+				CAConfigMapName:         cluster.Name + "-ca",
+				NodeSecretName:          cluster.Name + "-node-certs",
+				RootSQLClientSecretName: cluster.Name + "-client-certs",
+			},
+		},
+		Affinity:               cluster.Spec.Affinity,
+		NodeSelector:           cluster.Spec.NodeSelector,
+		Tolerations:            cluster.Spec.Tolerations,
+		TerminationGracePeriod: &metav1.Duration{Duration: time.Duration(cluster.Spec.TerminationGracePeriodSecs)},
```
prafull01, line 254: We need to update the cockroach-operator module to the latest so that its Spec has latest configuration.

scripts/migration/public/README.md
```diff
@@ -0,0 +1,118 @@
...
+```
+# Ensure that operator can't accidentally delete managed k8s objects.
+kubectl delete clusterrolebinding cockroach-operator-rolebinding
+
+# Delete public operator cr.
+kubectl delete crdbcluster $CRDBCLUSTER --cascade=orphan
+
+# Delete public operator resources and crd.
+kubectl delete -f https://raw.githubusercontent.com/cockroachdb/cockroach-operator/v2.17.0/install/crds.yaml
+kubectl delete -f https://raw.githubusercontent.com/cockroachdb/cockroach-operator/v2.17.0/install/operator.yaml
```
NishanthNalluri, line 61: I think we need to be careful here, if i set export NAMESPACE=cockroach-operator-system and the above commands will actually try to delete the same namespace, so the next operations won't work.

We need to specify to not set the NAMESPACE to  cockroach-operator-system

scripts/migration/public/README.md
```diff
@@ -0,0 +1,118 @@
...
+This guide will walk you through migrating a crdb cluster managed via the public operator to the crdb cloud operator. We assume you've created a cluster using the public operator. The goals of this process are to migrate without affecting cluster availability, and to preserve existing disks so that we don't have to replica data into empty volumes. Note that this process scales down the statefulset by one node before adding each operator-managed pod, so cluster capacity will be reduced by one node at times.
+
+Pre-requisite: Install the public operator and create an operator-managed cluster:
+
+```
+kubectl apply -f https://raw.githubusercontent.com/cockroachdb/cockroach-operator/v2.17.0/install/crds.yaml
+kubectl apply -f https://raw.githubusercontent.com/cockroachdb/cockroach-operator/v2.17.0/install/operator.yaml
+
+kubectl apply -f https://raw.githubusercontent.com/cockroachdb/cockroach-operator/v2.17.0/examples/example.yaml
+```
```
NishanthNalluri, line 12: As we mentioned `export NAMESPACE=default`, we can explicitly specify here to deploy example.yaml in the same namespace.

Following these commands will install crdbcluster in `cockroach-operator-system` namespace created while installing operator.

scripts/migration/public/README.md
prafull01: Actually this is fair point. If user have the crdbcluster CR in `cockroach-operator-system` then the statefulset is present in that namespace. This operation will delete the entire namespace and can cause data loss .

We should explicitly specify the resources we want to delete, rather than deleting from this link.

scripts/migration/public/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+```
+kubectl scale statefulset/$CRDBCLUSTER --replicas=2
+```
+
+Then create the crdbnode corresponding to the statefulset pod you just scaled down:
+
+```
+kubectl apply -f manifests/crdbnode-2.yaml
+```
```
NishanthNalluri, line 89: Can we have a note here, while repeating the process they need decrease replicas and use the respective manifest file.

prafull01 (approved):

PR #462, opened 2025-02-25
Manage Transparent Huge Pages for CRDB

MattWhelan:
THP is unhelpful for databases, in general, because memory access tends to be non-contiguous, and the use of THP can make performance worse when it requires a bunch of page defragmentation to load and map a huge page.

We disable it by default, but allow the chart user to modify the policy.

closes https://cockroachlabs.atlassian.net/browse/TREQ-888


build/templates/values.yaml
```diff
@@ -710,3 +710,15 @@ iap:
   # Create Google Cloud OAuth credentials and set client id and secret
   # clientId:
   # clientSecret:
+
+# Transparent Huge Page control
+# See https://www.cockroachlabs.com/docs/v25.1/recommended-production-settings.html#software
+thp:
```
pritesh-lahoti, line 716: We could also document this field under `build/templates/README.md`.

pritesh-lahoti (approved):

MattWhelan:
Apparently that sysctl isn't namespaced, so we can't set it this way. However, it looks like `GODEBUG=disablethp=1` may do the trick without changes to the Helm chart or code.

PR #463, opened 2025-02-28
Use the GODEBUG env var to disable Transparent Huge Pages

MattWhelan:
THP is unhelpful for databases, in general, because memory access tends to be non-contiguous, and the use of THP can make performance worse when it requires a bunch of page defragmentation to load and map a huge page.

We disable it by default, using the GODEBUG environment variable.

Adds a generic facility for managing GODEBUG clauses to the Helm chart.

closes https://cockroachlabs.atlassian.net/browse/TREQ-888


cockroachdb/templates/statefulset.yaml
```diff
@@ -260,6 +260,15 @@ spec:
               value: {{ template "cockroachdb.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
             - name: COCKROACH_CHANNEL
               value: kubernetes-helm
+            {{- define "godebugList" -}}
+            {{- $godebugList := list -}}
+            {{- range $key, $value := .Values.godebug }}
+              {{- $godebugList = append $godebugList (printf "%s=%s" $key $value) -}}
+            {{- end }}
+            {{- join "," $godebugList -}}
+            {{- end }}
```
pritesh-lahoti, line 269: This could move to `cockroachdb/templates/_helpers.tpl`.

pritesh-lahoti (approved):
Thank you for the changes!

PR #464, opened 2025-03-03
Cloud Operator support in helm chart for all the options of SHCM operator

prafull01:
- Added support for adding various fields for cockroach cloud operator.
- Added migration tooling which actually reads from the custom resource yaml and converts it into the values.yaml file to be installed as helm chart.

### DIFF
--- a/build/templates/values.yaml
+++ b/build/templates/values.yaml
@@ -716,13 +716,14 @@ operator:
   enabled: true
   # Default values for the cluster chart.
   image:
-    repository: cockroachdb/cockroach
+    name: cockroachdb/cockroach:v24.3.1
     pullPolicy: IfNotPresent
-    # Overrides the image tag whose default is the cluster chart's appVersion.
-    tag: ""
+    pullSecret: ""
 
   nameOverride: ""
   fullnameOverride: ""
+
+  # tlsEnabled: true
 
   # A map of CRDB cluster settings.
   # See https://www.cockroachlabs.com/docs/stable/cluster-settings.html
@@ -740,6 +741,10 @@ operator:
       # stderr:
         # channels: [health, dev]
         # filter: INFO
+
+  # loggingConfigMapName define the config map which contains log configuration used to send the logs through the
+  # proper channels in the cockroachdb.
+  loggingConfigMapName: ""
 
   # We usually recommend not to specify default resources and to leave this as a conscious
   # choice for the user. This also increases chances charts run on environments with little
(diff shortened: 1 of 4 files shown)


build/templates/values.yaml
```diff
@@ -716,14 +716,15 @@ operator:
   enabled: true
   # Default values for the cluster chart.
   image:
-    repository: cockroachdb/cockroach
+    name: cockroachdb/cockroach:v24.3.1
```
pritesh-lahoti, line 719: nit: We could also use v25.1.

cockroachdb/templates/crdb.yaml
```diff
@@ -25,32 +28,63 @@ spec:
...
           {{- /* Note: defaults should match secrets and configmaps created by the self-signer job. */}}
-          clientCaConfigMapName: {{ .Values.operator.certificates.externalCertificates.clientCaConfigMapName | default (printf "%s-ca-secret-crt" (include "cockroachdb.fullname" .)) }}
-          nodeCaConfigMapName: {{ .Values.operator.certificates.externalCertificates.nodeCaConfigMapName | default (printf "%s-ca-secret-crt" (include "cockroachdb.fullname" .)) }}
-          httpSecretName: {{ .Values.operator.certificates.externalCertificates.httpSecretName | default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
-          nodeClientSecretName: {{ .Values.operator.certificates.externalCertificates.nodeClientSecretName | default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
-          nodeSecretName: {{ .Values.operator.certificates.externalCertificates.nodeSecretName | default (printf "%s-node-secret" (include "cockroachdb.fullname" .)) }}
-          rootSqlClientSecretName: {{ .Values.operator.certificates.externalCertificates.rootSqlClientSecretName | default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
```
pritesh-lahoti, line 37: Was there any motivation behind having `externalCertificates.isEmpty`? Isn't the current change solving it already?

build/templates/values.yaml
```diff
@@ -716,14 +716,15 @@ operator:
...
     pullPolicy: IfNotPresent
-    # Overrides the image tag whose default is the cluster chart's appVersion.
-    tag: ""
+    pullSecret: ""
 
   nameOverride: ""
   fullnameOverride: ""
 
+  # tlsEnabled: true
```
pritesh-lahoti, line 726: Won't it make sense to derive this from `tls.enabled` config that exists already?

cockroachdb/templates/crdb.yaml
prafull01: Yes, we define default certificates when values.yaml has empty certificates. However, if the user provides their own certificates under external.certificates, there could be a mix of user-provided and default certificates. This is not valid and may cause issues during CockroachDB startup.

To prevent this, we only generate default certificates when external.certificates is empty. If the user provides any certificates, they are responsible for supplying all required certificates.

This user provided certs will come in picture while in migration of public operator.

build/templates/values.yaml
prafull01: I can do it right now but when we migrate this into separate chart then this field has to be introduced.

cockroachdb/templates/crdb.yaml
pritesh-lahoti: `externalCertificates.isEmpty` is being evaluated to `false` even if one of those certificates is provided.
So, I was wondering if the original change made sense, as it fallsback to using the config-map.

build/templates/values.yaml
pritesh-lahoti: Could this be a helper that determines whether TLS is enabled? Once we introduce subcharts, we could only update the helper logic.

prafull01: I think that will complicate things. I can use `tls.enabled` for now. When we come to subchart anyway we have to refactor many fields.

pritesh-lahoti: Sure, let's revisit this later.

PR #468, opened 2025-03-06
Added Operator charts, modified cockroachdb charts to support cc-operator

NishanthNalluri:
- Added CC-Operator Charts 
- Support to release in custom namespace
- CockroachDB charts have been updated to integrate with the cc-operator

Prev PR's: #456 #455

PR #470, opened 2025-03-06
helm: add documentation to deploy with v2 operator

NishanthNalluri:
- This PR includes documentation for deploying single and multi region CockroachDB Cluster through modified CockroachDB charts and CC-Operator Charts.
- Captures Steps to Perform for Different UseCases(Operations) via Helm on CockorachDB Cluster.
- Added new field `timestamp` in `cockroachdb/values.yaml` whose value is referred in pod annotation `helm.sh/restartedAt`, this is added for rolling restarts of CockroachDB pods without any other configuration changes.
- Update Operator image to include recent changes made to cc-operator
- Made default value of `operator.enabled` to `false`

PR #474, opened 2025-03-11
[CRDB-48127] Helm(V2) E2E Tests

NishanthNalluri:
- Added `dev-multi-cluster.sh` to setup infra for single region and multi region clusters.
- Added Make commands in `Makefile` to support infra setup and e2e tests run for single and multi region.
- Networking config needed for e2e tests are added under `e2e/calico` and `e2e/coredns`.
- Added e2e tests under `e2e/operator/` for single and multi region use cases.

Previous PR's - [#468] [#457] [#456]

JIRA - [CRDB-48127]

Parent - [CRDB-43233]


jmcarp (commented):
We're vendoring a lot of private code here. I don't understand yet what all of it is for, so I'll leave questions on the diff. We should also make sure it's acceptable to paste this code here, and we may want to figure out a way to move it to a separate library that we can import here.

Makefile
```diff
@@ -9,6 +9,9 @@ ifeq ($(UNAME_S),Linux)
   JQ_BIN ?= https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
   OPM_TAR ?= https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest-4.8/opm-linux-4.8.57.tar.gz
   OPM_BIN ?= opm
+  CILIUM_TAR ?= https://github.com/cilium/cilium-cli/releases/download/v0.16.24/cilium-linux-amd64.tar.gz
+  CILIUM_BIN ?= cilium
+  KIND_BIN ?=  https://kind.sigs.k8s.io/dl/v0.27.0/kind-linux-amd64
```
jmcarp, line 14: Are we using kind here? I thought we were running k8s via k3d/k3s.

Makefile
```diff
@@ -9,6 +9,9 @@ ifeq ($(UNAME_S),Linux)
   JQ_BIN ?= https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64
   OPM_TAR ?= https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/latest-4.8/opm-linux-4.8.57.tar.gz
   OPM_BIN ?= opm
+  CILIUM_TAR ?= https://github.com/cilium/cilium-cli/releases/download/v0.16.24/cilium-linux-amd64.tar.gz
```
jmcarp, line 12: Why do these tests need cilium?

build/templates/values.yaml
```diff
@@ -4,6 +4,9 @@ nameOverride: ""
 # Override the resource names created by this chart which originally is generated using release and chart name.
 fullnameOverride: ""
 
+# Sets timestamp annotation for CockroachDB pods
+timestamp: "1634572800"
```
jmcarp, line 8: What does this do?

build/templates/values.yaml
```diff
@@ -733,11 +736,6 @@ operator:
   # See https://www.cockroachlabs.com/docs/stable/cluster-settings.html
   clusterSettings: ~
 
-    # Regions controls the number of CockroachDB nodes that are deployed per region.
```
jmcarp, line 736: Why delete this?

tests/e2e/calico/calico.go
```diff
@@ -0,0 +1,293 @@
+package calico
```
jmcarp, line 1: Why do we need this code for these tests? Can we add a note about why we're vendoring it, and talk to cloud platform about extracting it to a separate repo if we need to use it in multiple repos?

tests/e2e/operator/region.go
```diff
@@ -0,0 +1,606 @@
...
+	rawConfig.CurrentContext = cluster
+
+	// Setup kubectl options for this cluster.
+	kubectlOptions := k8s.NewKubectlOptions(cluster, kubeConfig, r.Namespace[cluster])
+
+	// Create namespace.
+	k8s.CreateNamespace(t, kubectlOptions, r.Namespace[cluster])
+
+	// Create CA.
+	cmd := shell.Command{
```
jmcarp, line 302: Does this mean we're generating a new ca cert for each region in a multi-region cluster?

tests/e2e/operator/region.go
```diff
@@ -0,0 +1,606 @@
...
+
+		kubectlOptions := k8s.NewKubectlOptions(cluster, kubeConfig, r.Namespace[cluster])
+
+		pods := k8s.ListPods(t, kubectlOptions, metav1.ListOptions{
+			LabelSelector: LabelSelector,
+		})
+		require.NotEmpty(t, pods)
+
+		// Execute SQL query to verify regions
+		stdout, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
```
jmcarp, line 461: Are we also checking that pods in one region can connect to pods in the other region? Does `cockroach node status` see the right number of pods in each region, etc?

tests/k3d/dev-multi-cluster.sh
```diff
@@ -0,0 +1,252 @@
...
+    local k3d_args=(
+        "${K3D_PATH}" cluster create "${cluster_name}"
+#       --subnet ${subnet}
+        --no-lb
+#       --servers-memory 2GB
+       --k3s-arg "--service-cidr=${svc_cidr}@server:*"
+       --k3s-arg "--cluster-cidr=${cluster_cidr}@server:*"
+        --k3s-arg "--disable=traefik@server:*"
+        --k3s-arg "--disable-network-policy@server:*"
+#       --k3s-arg "--disable=servicelb@server:0"
```
jmcarp, line 117: What are these commented lines?

build/templates/values.yaml
NishanthNalluri: This is added as one of the annotation to the pods, so rolling restart can be done without changes to other config

More info on this here #470

Makefile
NishanthNalluri: Initially i was trying with cilium for multi region networking to make it work(unlike our gke setup only changes to coredns doesn't work with k3d setup), so i added cilium installation step in the script as well

Makefile
NishanthNalluri: Yeah while testing locally i didn't had luck creating more than 2 k3d clusters locally (3rd one is always getting stuck while creation), so i switched to kind to see if i can spin more than 2 clusters.

Anyway i feel we can remove these binary references in make commands

build/templates/values.yaml
NishanthNalluri: This was the redundant part, it already exists in other place under operator section

tests/k3d/dev-multi-cluster.sh
NishanthNalluri: These are the additional args you can pass while creating k3d clusters, so i just added few of them as commented args which can be useful sometimes while testing locally

tests/e2e/operator/region.go
NishanthNalluri: missed adding a condition for multi-region, will update.

tests/e2e/operator/region.go
NishanthNalluri: I haven't tried it, i was assuming it will show all the regions of connected pods through this sql query.
But i can verify this and add one more test if it works.

Makefile
jmcarp: Ok, if we're not using cilium in this PR, can we remove it?

tests/e2e/operator/region.go
jmcarp: If each region of the multi-region cluster is using a different ca cert, the cluster wouldn't function, right? Did this work in local testing?

NishanthNalluri: Yeah that's the same thing i mentioned if its multi region i will be using same CA for all other regions, while refactoring code i missed that condition (that's the reason my multi-region tests failed for this build)

tests/e2e/coredns/coredns.go
```diff
@@ -0,0 +1,434 @@
...
+
+		// The server values which will eventually find their way into the
+		// Corefile.
+		namespace := cluster.Namespace
+		var servers []string
+		if namespace != "" {
+			servers = append(
+				servers,
+				fmt.Sprintf("%s.svc.%s", namespace, clusterDomain),
+				fmt.Sprintf("%s.svc.%s", "monitoring-"+namespace, clusterDomain),
```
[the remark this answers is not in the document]
NishanthNalluri: i remember you mentioned this is not needed in our internal doc, will remove this.

PR #479, opened 2025-03-13
Fix: target port as sql for port 26257

NishanthNalluri:
This PR fixes the target port as `sql` for 26257 for `cockroach-service` as it should match with port exposed by pods.


pritesh-lahoti (approved):
LGTM, but I'm wondering if we should get rid of the port name from the values, as the operator hardcodes those strings and we don't want customers to change it here. Can be taken up separately.

PR #486, opened 2025-03-19
Add storageClassName field in the dataStore

prafull01:


build/templates/values.yaml
```diff
@@ -770,6 +770,9 @@ operator:
           requests:
             storage: 10Gi
         volumeMode: Filesystem
+      # storageClassName is the name of the StorageClass required by the claim.
+      # If not provided, it will use the default storage class of the cluster.
+      # storageClassName: ""
```
pritesh-lahoti, line 775:
```suggestion
        volumeMode: Filesystem
        # storageClassName is the name of the StorageClass required by the claim.
        # If not provided, it will use the default storage class of the cluster.
        # storageClassName: ""
```

pritesh-lahoti (approved):
Pre-approving.

PR #496, opened 2025-03-28
move flags within spec.template.spec in crdb.yaml

pritesh-lahoti:
Move flags from within `spec` to `spec.template.spec` to accommodate the corresponding changes in the operator.
Removed dangling `dataStore` entry within spec.


NishanthNalluri (approved):
LGTM

PR #497, opened 2025-04-03
Add changes for migration from public helm chart to cockroachdb enterprise operator

prafull01:
1. Added proper CLI structure to run the migration-helpers for operator and helm.
```
It generates the required manifest to migrate to the CockroachDB Enterprise Operator.

Usage:
  migration-helper build-manifest [command]

Available Commands:
  helm        Generate migration manifests for Cockroach Enterprise Operator from Official CockroachDB Helm chart(https://artifacthub.io/packages/helm/cockroachdb/cockroachdb)
  operator    Generate migration manifests for cockroach-enterprise-operator from cockroach-operator(https://github.com/cockroachdb/cockroach-operator)

Flags:
      --cloud-provider string   name of cloud provider
      --cloud-region string     name of cloud provider region
  -h, --help                    help for build-manifest
      --kubeconfig string       path to kubeconfig file (default "/Users/prafullladha/.kube/config")
      --output-dir string       manifest output directory (default "./manifests")

Use "migration-helper build-manifest [command] --help" for more information about a command.
```

2. Added unit tests for all the migration helpers.
3. Migrated existing migration changes to build manifest from cockroachdb operator.
4. Added support for the migration from the cockroachdb helm chart to CockroachDB Enterprise Operator.
5. Final output of the script:
```
prafullladha@Prafulls-MacBook-Pro helm-charts % bin/migration-helper build-manifest helm --statefulset-name $STS_NAME --namespace $NAMESPACE --cloud-provider $CLOUD_PROVIDER --cloud-region $REGION --output-dir ./manifests
✅ Migration manifests successfully generated.
📁 Output directory: ./manifests
📌 Next steps:
   1. Review the generated YAML files under the './manifests' directory.
   2. Follow the README.md under scripts/migration/helm directory
   3. Monitor the cluster to ensure a smooth transition.
```

Follow Up Tasks:
1. Implement certificate generation support within the existing migration-helper.
2. Integrate RBAC configuration into the migration-helper.
3. Enable execution of all migration helpers in a single run to simplify the user workflow.
4. Consolidate the migration steps for both the operator and Helm into a single README.md file.


cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/cockroachdb/helm-charts/pkg/migrate"
+	"github.com/spf13/cobra"
+)
+
+var (
+	crdbCluster     string
+	clusterManifest string
+)
+
+var buildManifestFromHelm = &cobra.Command{
```
jmcarp, line 13: Is this command specific to helm, or would it work with other statefulset-based crdb deployments? Is the idea that it works with statefulsets, but only supports options available in the helm chart?

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
```diff
@@ -0,0 +1,47 @@
...
+	Use:   "helm",
+	Short: "Generate migration manifests for CockroachDB Enterprise Operator from Helm chart",
+	Long:  "Generate manifests for migrating from the public CockroachDB Helm chart to the CockroachDB Enterprise Operator.",
+	RunE:  buildManifestFromCockroachDBHelmChart,
+}
+
+func init() {
+	buildManifestFromHelm.PersistentFlags().StringVar(&crdbCluster, "statefulset", "", "name of statefulset resource")
+	buildManifestFromHelm.PersistentFlags().StringVar(&namespace, "namespace", "default", "namespace of crdbcluster resource")
+	buildManifestFromHelm.PersistentFlags().StringVar(&clusterManifest, "statefulset-manifest", "", "path to statefulset manifest backup")
```
jmcarp, line 23: * Should it be an error to specify both "statefulset" and "statefulset-manifest"?
* I think we can give these flags more precise names—maybe "statefulset-name" and "statefulset-manifest".
* Do we explain how or why to use the backup option?

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
```diff
@@ -0,0 +1,47 @@
...
+var buildManifestFromHelm = &cobra.Command{
+	Use:   "helm",
+	Short: "Generate migration manifests for CockroachDB Enterprise Operator from Helm chart",
+	Long:  "Generate manifests for migrating from the public CockroachDB Helm chart to the CockroachDB Enterprise Operator.",
+	RunE:  buildManifestFromCockroachDBHelmChart,
+}
+
+func init() {
+	buildManifestFromHelm.PersistentFlags().StringVar(&crdbCluster, "statefulset", "", "name of statefulset resource")
+	buildManifestFromHelm.PersistentFlags().StringVar(&namespace, "namespace", "default", "namespace of crdbcluster resource")
```
jmcarp, line 22: Nit: s/crdbcluster resource/statefulset resource/

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
```diff
@@ -0,0 +1,47 @@
...
+
+var buildManifestFromHelm = &cobra.Command{
+	Use:   "helm",
+	Short: "Generate migration manifests for CockroachDB Enterprise Operator from Helm chart",
+	Long:  "Generate manifests for migrating from the public CockroachDB Helm chart to the CockroachDB Enterprise Operator.",
+	RunE:  buildManifestFromCockroachDBHelmChart,
+}
+
+func init() {
+	buildManifestFromHelm.PersistentFlags().StringVar(&crdbCluster, "statefulset", "", "name of statefulset resource")
```
jmcarp, line 21: Should we call this flag `statefulsetName` instead of `crdbCluster`?

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,454 @@
...
+	var startCmd string
+	var envVars = make(map[string]string)
+	var parsedInput = parsedMigrationInput{
+		tlsEnabled: true,
+	}
+
+	// In the public Helm chart, logging configuration is provided as a secret to the StatefulSet.
+	// However, in the Cockroach Enterprise Operator, it is supplied as a ConfigMap.
+	for _, vol := range sts.Spec.Template.Spec.Volumes {
+		if vol.Name == "log-config" {
```
jmcarp, line 269: Can we make special strings like "log-config" constants with descriptive names?

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,454 @@
...
+	parsedInput *parsedMigrationInput,
+	args []string,
+	envVars map[string]string) error {
+
+	flags := make(map[string]string)
+	// Regular expression to match flags (e.g., --advertise-host=something)
+	flagRegex := regexp.MustCompile(`--([\w-]+)=(.*)`)
+
+	for _, arg := range args {
+		arg = replaceEnvVars(arg, envVars) // Replace env variables
```
jmcarp, line 306: Why do we need to replace environment variables here? Can we copy environment variables to the operator manifest instead? Also, no need for a "Replace env variables" comment when we're calling a function called `replaceEnvVars`.

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
```diff
@@ -0,0 +1,47 @@
+package cockroachdb_enterprise_operator
```
jmcarp, line 1: We can delete `migrate/main.go` now, right?

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
prafull01: Point 1 &2, I will handle. 

I have added the manifest backup option as it was present in your PR. We are not deleting the sts until we are fully migrated, so we don't have to run it with manifest option ideally

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
prafull01: It is specifically for the statefulset created using helm chart of cockroachdb.

pkg/migrate/helpers.go
prafull01: I tested it before I have added env vars. So yes now we are adding the env vars so yes we can get rid of this func.

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,454 @@
...
+
+		case strings.HasPrefix(arg, "--insecure"):
+			parsedInput.tlsEnabled = false
+
+		// CockroachDB Enterprise Operator automatically adds "--logs" flag if it is not present. "--logtostderr" just
+		case strings.HasPrefix(arg, "--logtostderr"):
+			continue
+
+		default:
+			if matches := flagRegex.FindStringSubmatch(arg); len(matches) == 3 {
```
jmcarp, line 334: Does this handle flags that don't take an argument? Or do all flags match the regex?

pkg/migrate/helpers.go
```diff
@@ -0,0 +1,454 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create ConfigMap %s: %w", secretName, err)
+	}
+
+	fmt.Println("ConfigMap created successfully:", secretName)
+	return nil
+}
+
+// updatePublicService updates the "cockroachdb-public" service with separate sql and grpc ports.
+func updatePublicService(ctx context.Context, clientset kubernetes.Interface, namespace, name string) error {
```
jmcarp, line 407: * Remind me what the public svc is used for? Could updating it break client applications using the cluster?
* Since we're modifying an existing k8 resource, wdyt about backing up the original to disk before the update?

pkg/migrate/build-manifests.go
```diff
@@ -0,0 +1,267 @@
...
+	kubeconfig     string
+	outputDir      string
+	objectManifest string
+}
+
+// Option represents a functional option for configuring Manifest
+type Option func(*Manifest) error
+
+// NewMigration constructs a Manifest with required fields and functional options
+func NewMigration(cloudProvider, cloudRegion, kubeconfig, namespace, outputDir string, opts ...Option) (*Manifest, error) {
```
jmcarp, line 37: Nit: I would expect a constructor called `NewMigration` to return a result with type `Migration`.

jmcarp, line 37: Out of curiosity, what's the value of using a shared `Manifest` type between helm/operator sources? It looks like the two paths share almost zero code, and trying to abstract them leads us to use generic names like `object`, which is a bit confusing.

pkg/migrate/build-manifests.go
```diff
@@ -0,0 +1,267 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type Manifest struct {
+	cloudProvider  string
+	cloudRegion    string
+	object         string
```
jmcarp, line 26: How about calling this `objectName`, and renaming `objectManifest` to something like `objectPath` or `objectManifestPath`?

pkg/migrate/helpers.go
prafull01: Helm chart crdb statefulset runs sql and grpc both on 26257 port. The cockroachdb enterprise operator can not run both on 26257 port. Hence I am changing the grpc port to 26258 during migration. It will not break the sql connections but grpc commands needs to be run on 26258.

The public service is used to connect to cockroach sql or cockroach UI by users.

Yes. we can back it up to disk.

pkg/migrate/helpers.go
prafull01: All the flags matches this regex which have `key=value`. This consists of almost all the flags. The flags which doesn't follow this regex expression can not go into the flags map as it expects key value pairs.

pkg/migrate/helpers.go
prafull01: Instead of updating the service, I have generated the svc manifest. Added instruction in README.md to update the service.

pkg/migrate/build-manifests.go
prafull01: It is because the migration for operator and statefulset are very close.  Later we can only have one path for both migration. This is why I kept the same. In follow up tasks, I am going to simplify this.

cmd/migrate/cockroachdb-enterprise-operator/migrate-helm.go
prafull01: Removed the manifests option entirely

jmcarp (approved):
Thanks, this covers all my questions, and the follow-up items look good as well. LGTM!

PR #500, opened 2025-04-08
Fix password parsing in init job

ri-roee:
Caught this bug when deploying locally on minikube. The password getting set was the user name instead of the actual password being passed in.


prafull01 (approved):

ri-roee:
@prafull01 fixed the tests!

PR #502, opened 2025-04-22
Add functionalities and e2e tests for the migration

prafull01:
Added support of certificate migration inside migration-helper.
        Added support for rbac generation inside migration-helper.
        Added e2e tests for helm chart to cockroach enterprise operator.
        Added e2e tests for public operator to cockroach enterprise operator.


pkg/migrate/migrate_certs.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+
+		caCert = nodeSecret.CA()
+	} else {
+		caCert = secret.CA()
+	}
+
+	caKey = secret.CAKey()
+
+	// If we are using the operator to manage secrets then we need to store the CA cert in a
```
jmcarp, line 115: If this is only used to migrate to the enterprise operator, wouldn't `rc.OperatorManaged` always be `true`?

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	prepareForMigration(t, o.CrdbCluster.StatefulSetName, o.Namespace, o.CrdbCluster.CaSecret, "operator")
+	defer func() {
+		_ = os.RemoveAll(manifestsDirPath)
+	}()
+
+	o.UninstallOperator(t)
+	t.Log("Create the priority class crdb-critical which will be owned by the cockroachdb enterprise operator")
+	k8s.RunKubectl(t, kubectlOptions, "create", "priorityclass", "crdb-critical", "--value", "500000000")
+	defer k8s.RunKubectl(t, kubectlOptions, "delete", "priorityclass", "crdb-critical")
```
jmcarp, line 89: Do we need to wrap this in a function call?

pkg/migrate/migrate_certs.go
prafull01: Yes. I am setting the `rc.OperatorManaged` true initially itself. This func is derived from `func(r &genCert) LoadCASecret()` and instead of doing migration specific changes in the original source code, i created this new function.

I can get rid of this check: `if rc.CaSecret != "" && rc.OperatorManaged {}`

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
```diff
@@ -0,0 +1,125 @@
...
+	k8s.RunKubectl(t, kubectlOptions, "annotate", "service", fmt.Sprintf("%s-public", o.CrdbCluster.StatefulSetName), fmt.Sprintf("meta.helm.sh/release-name=%s", o.CrdbCluster.StatefulSetName), "--overwrite")
+	k8s.RunKubectl(t, kubectlOptions, "annotate", "service", fmt.Sprintf("%s-public", o.CrdbCluster.StatefulSetName), fmt.Sprintf("meta.helm.sh/release-namespace=%s", o.Namespace), "--overwrite")
+	k8s.RunKubectl(t, kubectlOptions, "label", "service", fmt.Sprintf("%s-public", o.CrdbCluster.StatefulSetName), fmt.Sprintf("app.kubernetes.io/managed-by=%s", "Helm"), "--overwrite")
+
+	o.HelmOptions = &helm.Options{
+		KubectlOptions: kubectlOptions,
+		ValuesFiles:    []string{filepath.Join(manifestsDirPath, "values.yaml")},
+		SetValues:      map[string]string{},
+	}
+	t.Log("helm install the cockroach enterprise operator")
```
jmcarp, line 113: Update this comment?

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
```diff
@@ -0,0 +1,125 @@
...
+	k8s.KubectlApply(t, kubectlOptions, filepath.Join(manifestsDirPath, "rbac.yaml"))
+
+	t.Log("Install the cockroachdb enterprise operator")
+	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	defer func() {
+		t.Log("Uninstall the cockroachdb enterprise operator")
+		operator.UninstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	}()
+
+	migratePodsToCrdbNodes(t, o.CrdbCluster.StatefulSetName, o.Namespace)
```
jmcarp, line 101: I'm wondering if we can make more assertions here. For example, should we assert that all crdb pods are ready after each one is updated? Assert that we don't see underreplicated ranges during the migration? Insert some data pre-migration and assert it's still present afterwards? One easy change might be to run `o.ValidateCRDB(t)` after each node is updated.

.github/workflows/ci.yaml
```diff
@@ -239,6 +239,31 @@ jobs:
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.22
+
+      - name: Run E2E Test
+        run: make test/e2e/migrate
+
+      - name: Clean up Docker resources
```
jmcarp, line 260: Out of curiosity, what's the benefit of this? Do images, volumes, etc live beyond the CI run or otherwise cause problems if not cleaned up? If this is useful, I think a comment about the purpose would help.

cmd/migrate/cockroachdb-enterprise-operator/migrate-certs.go
```diff
@@ -0,0 +1,92 @@
...
+The command supports customization of certificate durations and expiry windows for:
+- Node certificates (default: 1 year duration, 7 days expiry window)
+- Client certificates (default: 28 days duration, 2 days expiry window)
+
+The migration process ensures all certificates are properly configured for use with the CockroachDB Enterprise Operator.`,
+	RunE: migrateCertsForCockroachEnterpriseOperator,
+}
+
+func init() {
+	migrateCertsCmd.PersistentFlags().StringVar(&statefulSetName, "statefulset-name", "", "name of the cockroachdb statefulset resource")
```
jmcarp, line 45: Is this flag required?

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
prafull01: Yes, that is the plan. However this PR in itself became quite huge. So I am going to add more assertions in follow up PRs. I can do ValidateCRDB post every replica update in this PR.

cmd/migrate/cockroachdb-enterprise-operator/migrate-certs.go
prafull01: We need to know statefulset name to identify its public service name and join service name so that those can be added to SAN's while writing node cert

jmcarp: Can we mark it as required in that case?

pkg/migrate/helpers.go
```diff
@@ -403,6 +414,30 @@ func ConvertSecretToConfigMap(ctx context.Context, clientset kubernetes.Interfac
 	return nil
 }
 
+// moveConfigMapKey moves the "logging.yaml" key to "logs.yaml" in the ConfigMap.
+// This is a solution to support the migration from the public operator to the Cockroach Enterprise Operator.
+// The public operator uses "logging.yaml" and the Cockroach Enterprise Operator uses "logs.yaml".
+func moveConfigMapKey(ctx context.Context, clientset kubernetes.Interface, namespace, configMapName string) error {
```
jmcarp, line 420: If the user has to undo the migration and switch back to the public operator, we would need to revert this change, right? Do we even want to support the reverse migration? I don't think we need to make any changes at this point, but it's worth discussing separately.

jmcarp (approved):
I only have non-blocking questions. This LGTM!

pkg/migrate/helpers.go
prafull01: We are keeping the existing key in the configmap as well. So if user has to do reverse migration, this configmap will still be valid for him.

prafull01:
Merging the PR. I will handle all review comments in a follow up PR.

PR #507, opened 2025-04-29
Visus Sidecar config

MattWhelan:
First-class support for Visus as a CRDB Sidecar in the Helm chart.

Visus will start before the DB, and be restarted if it fails. K8s won't wait for visus to terminate, since the restartPolicy is set.

Also includes a separate headless Service def so the Prometheus scraper can address each node directly. Visus is not exposed outside the cluster.

### DIFF
--- a/tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
+++ b/tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
@@ -130,7 +130,7 @@ func (h *HelmChartToOperator) TestDefaultMigration(t *testing.T) {
 	require.Contains(t, err.Error(), "You are attempting to upgrade from a StatefulSet-based CockroachDB Helm chart to the CockroachDB Enterprise Operator.")
 
 	t.Log("Delete the StatefulSet as helm upgrade can proceed only if no StatefulSet is present")
-	k8s.RunKubectl(t, kubectlOptions, "delete", "statefulset", stsName)
+	k8s.RunKubectl(t, kubectlOptions, "delete", "statefulset", h.CrdbCluster.StatefulSetName)
 
 	helm.Upgrade(t, &helm.Options{
 		KubectlOptions: kubectlOptions,
--- a/tests/template/cockroachdb_helm_template_test.go
+++ b/tests/template/cockroachdb_helm_template_test.go
@@ -117,6 +117,36 @@ func TestHelmSelfCertSignerServiceAccount(t *testing.T) {
 	require.Contains(t, err.Error(), "Error: could not find template templates/serviceaccount-certSelfSigner.yaml in chart")
 }
 
+// TestHelmVisusSidecar verifies the visus sidecar yaml
+func TestHelmVisusSidecar(t *testing.T) {
+	t.Parallel()
+
+	// Setup the args. For this test, we will set the following input values:
+	options := &helm.Options{
+		KubectlOptions: k8s.NewKubectlOptions("", "", namespaceName),
+		SetValues: map[string]string{
+			"visus.enabled": "true",
+		},
+	}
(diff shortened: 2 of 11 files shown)


MattWhelan:
There seem to be test failures unrelated to my changes, possibly due to merge skew in the `master` branch. One seems to be a missed refactor that just fails to compile; I've addressed that I think. The others I don't really understand.

tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
```diff
@@ -130,7 +130,7 @@ func (h *HelmChartToOperator) TestDefaultMigration(t *testing.T) {
 	require.Contains(t, err.Error(), "You are attempting to upgrade from a StatefulSet-based CockroachDB Helm chart to the CockroachDB Enterprise Operator.")
 
 	t.Log("Delete the StatefulSet as helm upgrade can proceed only if no StatefulSet is present")
-	k8s.RunKubectl(t, kubectlOptions, "delete", "statefulset", stsName)
+	k8s.RunKubectl(t, kubectlOptions, "delete", "statefulset", h.CrdbCluster.StatefulSetName)
```
MattWhelan, line 133: `stsName` doesn't exist, but I see it was replaced recently. Merge skew?

pritesh-lahoti: Yeah, this has been fixed now. Please rebase.

pritesh-lahoti (commented):
Thank you for the changes, @MattWhelan . They look good to me.
Currently, we're working on a v2 offering for Helm charts that uses the Enterprise Operator for cluster management.

Do we intend to add the visus configuration for those charts? They reside under `cockroachdb-parent`.

MattWhelan:
> Thank you for the changes, @MattWhelan . They look good to me. Currently, we're working on a v2 offering for Helm charts that uses the Enterprise Operator for cluster management.
> 
> Do we intend to add the visus configuration for those charts? They reside under `cockroachdb-parent`.

Yeah, we'll want a good way to integrate Visus in that setup as well. Since Visus exposes per-node state, it needs some special treatment (that headless service) so that each instance is addressable. Other sidecars (like a log shipper/filter, or even push-based telemetry collectors) don't generally need that special treatment, since they don't expose a service.

tests/template/cockroachdb_helm_template_test.go
```diff
@@ -117,6 +117,36 @@ func TestHelmSelfCertSignerServiceAccount(t *testing.T) {
...
+	helm.UnmarshalK8SYaml(t, output, &sts)
+	require.Equal(t, "visus", sts.Spec.Template.Spec.InitContainers[1].Name)
+	require.Equal(t, []string{
+		"visus",
+		"start",
+		"--url",
+		"postgres://root@localhost:26257/defaultdb?application_name=visus&sslmode=verify-full&ssrootcert=/cockroach/client/ca.crt&sslcert=/cockroach/client/client.root.crt&sslkey=/cockroach/client/client.root.key",
+		"--visus-metrics",
+		"--bind-addr",
+		"localhost:8888",
```
MarcPaquette, line 145: Would `--bind-addr localhost:8888` be better as `--bind-addr 0.0.0.0:8888` to bind to all interfaces for external scraping?

MattWhelan: That's not how k8s networking works. Whatever addresses you bind, they're internal to the container. To expose a service in k8s, you do it via k8s networking constructs, like a k8s `Service`. If you look at `service.visus.yaml`, you'll see that. 

The service definition is intended to work with the Prometheus Operator, which will discover the service endpoints via their annotations, and wire them up to the scraper. The is also how the normal Cockroach metrics get into Prometheus.

PR #511, opened 2025-05-02
Add upgrade validation check to avoid accidental upgrades from statefulset based helm chart without migration.

prafull01:
We are keeping the same name of the chart for Cockroach Enterprise Operator. We would be upgrading the version to much higher version to indicate that we have added a significant change how cockroachdb cluster is managed.

To avoid any accidental upgrades without migration, I have added this pre-upgrade check. It checks if there is existing statefulset present with the same name. If yes, then we print this message to user to migrate and doesn't allow upgrade to happen.

### DIFF
--- a/cockroachdb-parent/charts/cockroachdb/templates/_helpers.tpl
+++ b/cockroachdb-parent/charts/cockroachdb/templates/_helpers.tpl
@@ -310,3 +310,15 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/name: {{ include "cockroachdb.clusterfullname" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "cockroachdb.isUpgradeAllowed" -}}
+{{- $stsName := include "cockroachdb.fullname" . -}}
+{{- $sts := lookup "apps/v1" "StatefulSet" .Release.Namespace $stsName -}}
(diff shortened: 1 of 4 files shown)


cockroachdb-parent/charts/cockroachdb/templates/_helpers.tpl
```diff
@@ -310,3 +310,15 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/name: {{ include "cockroachdb.clusterfullname" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 {{- end }}
+
+{{- define "cockroachdb.isUpgradeAllowed" -}}
+{{- $stsName := include "cockroachdb.fullname" . -}}
```
pritesh-lahoti, line 315: Ideally, this should be referring to the one in `cockroachdb/templates/_helpers.tpl`, but I believe it should work fine as long as the release/chart names are same.

pritesh-lahoti (approved):

cockroachdb-parent/charts/cockroachdb/templates/_helpers.tpl
prafull01: Yes. We are doing this validation as we want to keep the chart name same.

NishanthNalluri (approved):
LGTM

PR #513, opened 2025-05-02
Improved the migration e2e tests

prafull01:
Handled all review comments from #502


jmcarp (approved):
Thanks for the fixes. LGTM once tests are passing.

PR #514, opened 2025-05-12
Add template unit tests for the migration

prafull01:
- Add template unit tests for FromHelm
- Add template unit tests for FromOperator


pkg/migrate/build-manifests_test.go
```diff
@@ -0,0 +1,184 @@
+package migrate
```
pritesh-lahoti, line 1: Do we plan to add an `update` flag to update the golden file output automatically, in case of changes?

pkg/migrate/testdata/helmAllInput/cockroachdb-public.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
```
pritesh-lahoti, line 1: nit: should we nest this as `.../testdata/helm/allInputs/...` and `.../testdata/operator/allInputs/...`?
in case we want to add tests for other combinations going forward.

pritesh-lahoti: Slightly related - should we rename `scripts/migration/public` to `scripts/migration/operator`?

pritesh-lahoti (approved):

pkg/migrate/build-manifests_test.go
prafull01: In case of any change in migration logic which changes the output yaml files. These tests are supposed to fail and the dev who is changing the migration can verify that the output generated is correct and add same fields in the golden files.

What do you have in mind to update it automatically?

pkg/migrate/testdata/helm/allInput/crdbnode-0.golden.yaml
```diff
@@ -0,0 +1,101 @@
+apiVersion: crdb.cockroachlabs.com/v1alpha1
```
pritesh-lahoti, line 1: nit: I'm wondering if it's better to suffix these files as `.yaml.golden` instead.

pritesh-lahoti (approved):

pkg/migrate/build-manifests_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	if *updateGolden {
+		generatedData, err := os.ReadFile(generatedFile)
+		require.NoError(t, err)
+		err = os.WriteFile(goldenFile, generatedData, 0644)
+		require.NoError(t, err)
+		t.Logf("Updated golden file: %s", goldenFile)
+	}
+
+	assert.Equal(t, string(golden), string(generated), "Generated file does not match golden file")
```
pritesh-lahoti, line 194: "Generated file does not match golden file. Please run the test with the -update flag to update the golden files."

PR #519, opened 2025-05-20
Add templates for cockroachdb operator helm charts

prafull01:


cockroachdb-parent/Chart.lock
```diff
@@ -1,9 +1,9 @@
...
 - name: operator
   repository: file://charts/operator
-  version: 0.1.0
+  version: 20.0.0
 - name: cockroachdb
   repository: file://charts/cockroachdb
-  version: 16.0.6
-digest: sha256:4b8ca92ac993f95f3a04249c01a5c1c53271b7da61ba743ed587f879d42fb491
-generated: "2025-04-24T15:32:38.520476+05:30"
+  version: 20.0.0
```
pritesh-lahoti, line 7: I am now wondering if we could use `25.0.0-preview.1` to explicitly call out that this is a preview release (non-GA).
Once we GA, we could align it to CRDB version (25.3.0, etc.).

build/build.go
```diff
@@ -34,13 +34,39 @@ const (
 	chartsFileTemplate = "build/templates/Chart.yaml"
 	valuesFileTemplate = "build/templates/values.yaml"
 	readmeFileTemplate = "build/templates/README.md"
+
```
pritesh-lahoti, line 37: nit: would this hierarchy be more intuitive?

\- build
-- templates
--- cockroachdb
---- Chart.yaml
--- cockroachdb-parent
---- charts
----- cockroachdb
------ Chart.yaml
----- operator
------ Chart.yaml

cockroachdb-parent/Chart.lock
prafull01: I have set it to `25.0.0-preview` and whenever someone bumps it until GA, it will become `25.0.1-preview`

pritesh-lahoti: In that case, I'm thinking if we should align it to the current CRDB version itself.

Looking at https://github.com/cockroachdb/helm-charts/pulls?q=is%3Apr+is%3Aclosed+author%3Acockroach-teamcity, we always only increment forward, so we could just sync with the current CRDB version and name it as `25.1.6-preview`.

Once we GA, we just remove the `-preview` suffix.

build/build.go
```diff
@@ -193,6 +218,40 @@ func bumpVersion(chart versions, newCRDBVersion *semver.Version) (string, error)
...
+		}
+
+		// Construct the new version string
+		return fmt.Sprintf("%s+%s", baseVersionStr, newBuild), nil
+	}
+
+	// If the version includes a prerelease label like `-preview`, the IncrPatch function automatically removes it.
+	// For example, v25.0.0-preview becomes v25.0.0 after incrementing the patch.
+	// To retain the prerelease label, we now increment the patch twice and re-append the prerelease.
+	// This ensures that versions with a prerelease set maintain the label even after a patch increment.
```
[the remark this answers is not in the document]
prafull01: There are two scenarios here:
1. With `-preview` you bump the cockroachdb app version as well. In that case, we are going to bump the patch and add the preview prerelease to it. For example: `v25.1.6` is update to `v25.1.7`, then chart version will go from `v25.1.6-preview` to `v25.1.7-preview`
2. With `-preview` option you only do helm changes and it doesn't update cockroachdb app version in that case the patch is not update but build metadata is added. For example: `v25.1.6-preview` becomes `v25.1.6-preview+1`

I have added test cases for the same.

build/build_test.go
```diff
@@ -0,0 +1,185 @@
...
+	}{
+		{
+			name:         "bump major version",
+			chartVersion: "25.1.6",
+			appVersion:   "25.1.6",
+			newCRDB:      "26.0.0",
+			gotVersion:   "26.0.0",
+		},
+		{
+			name:         "bump minor version",
```
pritesh-lahoti, line 33:
```suggestion
			name:         "bump major version series",
```

build/build_test.go
```diff
@@ -0,0 +1,185 @@
...
+func TestBumpVersion_MajorOrMinorChange(t *testing.T) {
+	testcases := []struct {
+		name         string
+		chartVersion string
+		appVersion   string
+		newCRDB      string
+		gotVersion   string
+	}{
+		{
+			name:         "bump major version",
```
pritesh-lahoti, line 26:
```suggestion
			name:         "bump major version calendar year",
```

pritesh-lahoti (approved):
Thank you, @prafull01 !

PR #527, opened 2025-06-11
Update to cockroach version 25.2.1

rail:


cockroachdb/Chart.yaml
```diff
@@ -2,8 +2,8 @@
 apiVersion: v1
 name: cockroachdb
 home: https://www.cockroachlabs.com
-version: 17.0.0
-appVersion: 25.2.0
+version: 17.0.1
```
rail, line 5: @prafull01 Should this be in sync with the crdb version (25.2.1)?

prafull01: @rail Current helm chart version are not in sync with the CockroachDB versions. The new helm chart which installs the cockroach enterprise operator and cluster will be in sync with CockroachDB versions.

prafull01 (approved):

PR #531, opened 2025-06-26
Remove the nodeClientCerts from self signer and cert manager

prafull01:
We are not signing the node client certs in case of self signer and cert manager. Hence we are not providing those input.

### DIFF
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -287,7 +287,6 @@ func buildNodeSpecFromHelm(
 				NodeSecretName:          input.nodeSecretName,
 				RootSQLClientSecretName: input.clientSecretName,
 				HTTPSecretName:          input.clientSecretName,
-				NodeClientSecretName:    input.clientSecretName,
 			},
 		},
 		Affinity:                  sts.Spec.Template.Spec.Affinity,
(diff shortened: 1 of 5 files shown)


NishanthNalluri (approved):

cockroachdb-parent/charts/cockroachdb/templates/crdb.yaml
```diff
@@ -45,15 +45,13 @@ spec:
...
           httpSecretName: {{ default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
-          nodeClientSecretName: {{ default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
           nodeSecretName: {{ default (printf "%s-node-secret" (include "cockroachdb.fullname" .)) }}
           rootSqlClientSecretName: {{ default (printf "%s-client-secret" (include "cockroachdb.fullname" .)) }}
         {{- end }}
         {{- if .Values.cockroachdb.tls.certManager.enabled }}
         externalCertificates:
           caConfigMapName: {{ .Values.cockroachdb.tls.certManager.caConfigMap }}
           nodeSecretName: {{ .Values.cockroachdb.tls.certManager.nodeSecret }}
-          nodeClientSecretName: {{ .Values.cockroachdb.tls.certManager.clientRootSecret }}
```
pritesh-lahoti, line 56: Let's also remove it from the commented section in `values.yaml` and migration logic.

prafull01: I would like to keep it in commented section as it is still valid input for the cockroachdb-operator. It is just that we are not providing it as a part of self signer and cert manager as it is not required input.

prafull01: Removed it from migration

pritesh-lahoti (approved):

PR #541, opened 2025-07-24
[CRDB-51768] [CRDB-51638] Add localityMappings and remove join fields to/from crdb spec.

NishanthNalluri:
- `localityMappings` gives flexibility to configure `nodeLabel` and `localityLabel`.
- `localityLabels` is deprecated and localityMappings is the recommended way to configure
    for computing locality dynamically for each node.
- `join` needs to be passed via `startFlags`.
-  Bumped helm version and updated `CHANGELOG`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file.
 
+## [cockroachdb-parent-25.2.2-preview] - 2025-07-28
+### Added
+- `startFlags`, `podTemplate` fields for overriding CockroachDB start command and pod spec. 
+- `localityMappings` field to allow granular mapping of Kubernetes node label to CockroachDB node locality.
+
+### Changed
+- Removed the deprecated `flags` field; use `startFlags` instead.
+- Removed the `join` field; specify it using `startFlags`.
+
 ## [cockroachdb-17.0.0] - 2025-05-27
 ### Added
 - release: advance app version to v25.2.0
--- a/build/templates/cockroachdb-parent/charts/cockroachdb/values.yaml
+++ b/build/templates/cockroachdb-parent/charts/cockroachdb/values.yaml
@@ -168,9 +168,6 @@ cockroachdb:
     clusterSettings: ~
     # timestamp captures the annotation timestamp used for rolling restarts.
     timestamp: "2021-10-18T00:00:00Z"
-    # join captures the list of comma-separated addresses for joining an existing cluster.
-    # Each item should be a resolvable FQDN (with port if needed).
-    join: ""
     # resources captures the resource requests and limits for CockroachDB pods.
     resources: ~
       # limits:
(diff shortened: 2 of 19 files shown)


build/templates/cockroachdb-parent/charts/cockroachdb/values.yaml
```diff
@@ -408,8 +405,27 @@ cockroachdb:
...
+    # Each locality mapping captures the relationship between a node label and a locality label.
+    # For region and zone to be part of the locality, the labels (topology.kubernetes.io/region, topology.kubernetes.io/region) must be set on the nodes.
+    # For other labels, they must be set on the nodes and will be presented as key/value pairs to the node locality.
+    # Example:
+    # If localityMappings are provided as
+    # {
+    #   {nodeLabel: "topology.kubernetes.io/region", localityLabel: "region"},
+    #   {nodeLabel: "topology.kubernetes.io/zone", localityLabel: "zone"},
+    #   {nodeLabel: "example.datacenter.locality", localityLabel: "dc"}
+    # }
```
pritesh-lahoti: nit/optional - should we also add this as a commented value for the field?

```
localityMappings: {}
  # - nodeLabel: "topology.kubernetes.io/region"
  #   localityLabel: "region"
  # - nodeLabel: "topology.kubernetes.io/zone"
  #   localityLabel: "zone"
  # - nodeLabel: "example.datacenter.locality"
  #   localityLabel: "dc"
```

prafull01:
> @NishanthNalluri Won't we need to make changes on the migrations side too (for `join`)? cc - @prafull01

@pritesh-lahoti Yes. In last PR also I explained nishant, that we need to do changes in migration. This is the reason migration tests are failing

NishanthNalluri:
> > @NishanthNalluri Won't we need to make changes on the migrations side too (for `join`)? cc - @prafull01
> 
> @pritesh-lahoti Yes. In last PR also I explained nishant, that we need to do changes in migration. This is the reason migration tests are failing

I have actually done those changes, the migration tests are failing because of webhook validation as both flags and startFlags are set but i made sure flags is removed entirely from migration code or else unit tests would have failed.

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+## [cockroachdb-parent-25.2.2-preview] - 2025-07-28
+### Added
+- `startFlags`, `podTemplate` fields for overriding CockroachDB start command and pod spec. 
+- `localityMappings` field to allow granular mapping of Kubernetes node label to CockroachDB node locality.
+- `noCloudPrefixedLocalities` field to disable automatic prefixing of locality labels with "cloud provider".
+
+### Changed
+- Removed deprecated `flags` field.
+- Removed `join` field.
```
pritesh-lahoti: nit: we could also mention here on the new fields to be used instead.

build/templates/cockroachdb-parent/charts/cockroachdb/values.yaml
```diff
@@ -408,8 +400,37 @@ cockroachdb:
...
+      # - nodeLabel: "topology.kubernetes.io/region"
+      #   localityLabel: "region"
+      # - nodeLabel: "topology.kubernetes.io/zone"
+      #   localityLabel: "zone"
+      # - nodeLabel: "example.datacenter.locality"
+      #   localityLabel: "dc"
+    # noCloudPrefixedLocalities indicates whether CockroachDB node locality flag is prefixed with its cloud provider's short code.
+    # If cloudProvider is set to "aws", each locality label will be prefixed with "aws" (e.g., "region=aws-us-east-1,zone=aws-us-east-1a").
+    # If set to false or if cloudProvider is an empty string, the locality flag will not be prefixed with the cloud provider's short code (e.g., "region=us-east-1,zone=us-east-1a").
+    noCloudPrefixedLocalities: true
```
pritesh-lahoti: nit: from a Helm or self-hosted customer's point-of-view, is 'true' the ideal default here?
Also, should we rename the field to `cloudPrefixedLocalities` within Helm at least?

PR #551, opened 2025-08-26
[CC-33398] cockroachdb/templates: Add support for loggingConfigVars field.

NishanthNalluri:


pritesh-lahoti (approved):
@NishanthNalluri - we also need to update the operator image version, right?

NishanthNalluri:
> @NishanthNalluri - we also need to update the operator image version, right?

Yeah, missed this, updated now.

PR #556, opened 2025-09-09
[CRDB-51649] Remove the dependency of the cockroach-operator test helper functions

prafull01:
We were importing many test helpers from the https://github.com/cockroachdb/cockroach-operator. This PR removes that dependency by moving test helper functions in this repo itself.


pkg/migrate/build-manifests_test.go
```diff
@@ -5,10 +5,11 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
-	"sigs.k8s.io/yaml"
 	"strconv"
 	"testing"
 
+	"sigs.k8s.io/yaml"
```
pritesh-lahoti: nit: this import can be merged with the imports in the section below.

PR #557, opened 2025-09-10
[ CC-32596 ] Add WAL failover custom path support for helm chart of cockroachdb operator deployments and handle the migration changes

shreyaskm623:
This PR adds support for -
 - Custom path configuration via `CrdbWalFailoverSpec.Path` field
 - Handling migration of WAL failover settings from Helm StatefulSets to the operator based deployments.

### DIFF
--- a/cockroachdb-parent/charts/cockroachdb/values.yaml
+++ b/cockroachdb-parent/charts/cockroachdb/values.yaml
@@ -277,6 +277,15 @@ cockroachdb:
       # If not set, the default provisioner will be chosen (gp2 on AWS, standard on GKE).
       # storageClassName: ""
       #
+      # path is the absolute mount path inside the CockroachDB container to use for WAL failover side disk.
+      # Examples: 
+      #   - /cockroach/cockroach-failover
+      #   - /cockroach/custom/path
+      # IMPORTANT: Set it only for fresh deployments. Changing it on upgrades might cause issues with the deployment.
+      # TODO: Add validation in the operator to prevent changing this value on upgrades.
+      #
+      # path: "/cockroach/cockroach-failover"
+      #
     # podLabels captures additional labels to apply to CockroachDB pods.
     podLabels:
       app.kubernetes.io/component: cockroachdb
--- a/pkg/migrate/helpers.go
+++ b/pkg/migrate/helpers.go
@@ -372,56 +372,90 @@ func buildHelmValuesFromHelm(
 		}
 	}
 
+	// Derive walFailoverSpec from --wal-failover flag if present.
+	var walSpec map[string]interface{}
+	if input.startFlags != nil {
+		for _, f := range input.startFlags.Upsert {
+			if strings.HasPrefix(f, "--wal-failover=") {
+				val := strings.TrimPrefix(f, "--wal-failover=")
+				if val == "disabled" {
+					walSpec = map[string]interface{}{
+						"status": "disable",
+					}
+				} else if strings.HasPrefix(val, "path=") {
+					path := strings.TrimPrefix(val, "path=")
+					walSpec = map[string]interface{}{
+						"status": "enable",
+						"path":   path,
+					}
+				} else if val == "among-stores" {
+					// among-stores is enabled but has no custom path
+					walSpec = map[string]interface{}{
+						"status": "enable",
+					}
+				}
+			}
+		}
+	}
+	crdbCluster := map[string]interface{}{
+		"image": map[string]interface{}{
+			"name": sts.Spec.Template.Spec.Containers[0].Image,
+		},
+		"localityLabels": input.localityLabels,
+		"podLabels":      sts.Spec.Template.Labels,
+		"podAnnotations": sts.Spec.Template.Annotations,
+		"resources":      sts.Spec.Template.Spec.Containers[0].Resources,
+		"startFlags":     input.startFlags,
+		"regions": []map[string]interface{}{
+			{
+				"namespace":     namespace,
+				"cloudProvider": cloudProvider,
+				"code":          cloudRegion,
+				"nodes":         sts.Spec.Replicas,
+				"domain":        "",
+			},
+		},
+		"dataStore": map[string]interface{}{
+			"volumeClaimTemplate": map[string]interface{}{
+				"metadata": map[string]interface{}{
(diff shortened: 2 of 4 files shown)


cockroachdb-parent/charts/cockroachdb/values.yaml
```diff
@@ -277,6 +282,15 @@ cockroachdb:
       # If not set, the default provisioner will be chosen (gp2 on AWS, standard on GKE).
```
prafull01, line 282: As per our discussion it is not true. When left empty in CockroachDB operator it doesn't pickup the default storage class. So we have to take an input from user if enabled.

pkg/migrate/helpers.go
```diff
@@ -272,13 +272,98 @@ func buildIngressValue(cluster publicv1.CrdbCluster) map[string]interface{} {
...
+					failoverStorageClass = actualStorageClass
+				}
+
+				if val == "disabled" {
+					walSpec = &v1alpha1.CrdbWalFailoverSpec{
+						Status: "disable",
+					}
+				} else if strings.HasPrefix(val, "path=") {
+					path := strings.TrimPrefix(val, "path=")
+					walSpec = &v1alpha1.CrdbWalFailoverSpec{
```
prafull01: This func needs to be refactored to cover edge cases:

1. At line 315, as soon as you get val, the `val` can have these possible values: `disabled`, `among-stores` or. `path=anypath`.
2. As soon as you get val, you check if it is disabled, then you set the`walSpec.Status: disable`. You don't need to find the PVC after that as it is disabled and the PVC will not be there.
3. if it is among-store, you should print out saying the multi store is not supported in new operator.
4. if it is path, then you start getting the details from PVC.

At line 320-330, it is not needed. We have a function getPVCStorageClassNameForNode which gets the SC of the PVC. We should rename that func getWalFailoverPVCDetails and this func should return everything we need from that PVC i.e. Name, size and SC.

pkg/migrate/build-manifests.go
```diff
@@ -182,7 +184,12 @@ func (m *Manifest) FromHelmChart() error {
 			return errors.Newf("pod %s isn't scheduled to a node", podName)
 		}
 
-		nodeSpec := buildNodeSpecFromHelm(sts, pod.Spec.NodeName, input)
+		// Fetch the walFailoverSpec if applicable
+		walSpec := getWalFailoverSpec(ctx, m.clientset, sts, pod.Spec.NodeName, nodeIdx, input)
```
prafull01: We should add a field `walFailoverSpec` in the `parsedMigrationInput` struct. 
Rename the `func buildWalFailoverSpec` and keep the walFailoverSpec in the input.
This way we don't need to pass it separately in buildNodeSpecFromHelm.

prafull01 (commented):
Based on suggested changes, the test cases will also change.

tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
```diff
@@ -151,6 +156,49 @@ func (h *HelmChartToOperator) TestDefaultMigration(t *testing.T) {
 
 	h.ValidateExistingData = true
 	h.ValidateCRDB(t)
+
+	// 	t.Log("Validating WAL failover configuration after migration")
```
pritesh-lahoti, line 161: nit: is this commented section expected?

cockroachdb-parent/charts/cockroachdb/values.yaml
```diff
@@ -277,6 +282,15 @@ cockroachdb:
...
       # storageClassName: ""
       #
+      # path is the absolute mount path inside the CockroachDB container to use for WAL failover side disk.
+      # Examples: 
+      #   - /cockroach/cockroach-failover
+      #   - /cockroach/custom/path
+      # IMPORTANT: Set it only for fresh deployments. Changing it on upgrades might cause issues with the deployment.
+      # TODO: Add validation in the operator to prevent changing this value on upgrades.
```
pritesh-lahoti, line 290: nit: we could instead add this TODO in `pkg/upstream/cockroach-operator/api/v1alpha1/crdbcluster_types.go` next to the `Path` field.

pkg/migrate/helpers.go
```diff
@@ -272,13 +279,105 @@ func buildIngressValue(cluster publicv1.CrdbCluster) map[string]interface{} {
...
+		size:             failoverSize,
+		storageClassName: failoverStorageClass,
+	}, nil
+}
+
+func buildWalFailoverSpec(ctx context.Context, clientset kubernetes.Interface, sts *appsv1.StatefulSet, nodeName string, nodeIdx int32, input *parsedMigrationInput) {
+	// Derive walFailoverSpec from --wal-failover flag if present and store it in input.
+	if input.startFlags != nil {
+		for _, f := range input.startFlags.Upsert {
+			if strings.HasPrefix(f, "--wal-failover=") {
```
pritesh-lahoti, line 387: nit: we could reduce the overall cognitive complexity of this function.

if !strings.HasPrefix(f, "--wal-failover=") {
  continue
}
val := strings.TrimPrefix(f, "--wal-failover=")
...

cockroachdb-parent/charts/operator/values.yaml
```diff
@@ -10,7 +10,7 @@ image:
   # pullPolicy specifies the image pull policy.
   pullPolicy: IfNotPresent
   # tag is the image tag.
-  tag: "6f62639e9fee99d99b0387a9dccda84daa1a489b592b008f2f354ec57eae09ac"
+  tag: "496a3d0b4b4a38637d97299aec52646c41eb32380c56b425376c8dd262c8b882"
```
prafull01, line 13: Your Image is not correct, that is why test is not working. Check the registry and repository as well

cockroachdb-parent/charts/cockroachdb/values.yaml
shreyaskm623: @prafull01 In case of a fresh deployment - if the storageClassName is not provided , it picks up the default provisioner , only in case of migration , since the storageClassName would already be present , we not giving the className in the spec would cause problems. So populating the storageClassName is important during migration and not mandatory for fresh deployments.

PR #563, opened 2025-10-14
[CC-54713] Migration should use podTemplate instead of deprecated fields

prafull01:


pkg/migrate/helpers.go
```diff
@@ -113,12 +112,49 @@ func yamlToDisk(path string, data []any) error {
...
+				},
+				ServiceAccountName:            cluster.Name,
+				Affinity:                      sts.Spec.Template.Spec.Affinity,
+				NodeSelector:                  sts.Spec.Template.Spec.NodeSelector,
+				Tolerations:                   sts.Spec.Template.Spec.Tolerations,
+				TerminationGracePeriodSeconds: &cluster.Spec.TerminationGracePeriodSecs,
+				TopologySpreadConstraints:     sts.Spec.Template.Spec.TopologySpreadConstraints,
+			},
+		},
+		// PodLabels:      sts.Spec.Template.Labels,
```
NishanthNalluri: nit: Do we need this commented code?

PR #565, opened 2025-10-15
[ CC-32596 ] Add WAL failover custom path support for helm chart migration

shreyaskm623:


build/templates/cockroachdb-parent/charts/operator/values.yaml
```diff
@@ -3,13 +3,13 @@
 # image captures the container image settings for Operator pods.
 image:
   # registry is the container registry where the image is stored.
-  registry: "us-docker.pkg.dev/releases-prod/self-hosted"
+  registry: "us-docker.pkg.dev/cockroach-cloud-images/development"
```
NishanthNalluri: We should not use dev image.

pkg/migrate/helpers_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	`strings`
```
NishanthNalluri:
```suggestion
	"strings"
```

pkg/migrate/helpers.go
```diff
@@ -369,13 +376,105 @@ func detectPCRFromInitJob(clientset kubernetes.Interface, stsName, namespace str
...
+	}
+
+	return &walFailoverPVCDetails{
+		name:             failoverPVCName,
+		size:             failoverSize,
+		storageClassName: failoverStorageClass,
+	}, nil
+}
+
+func BuildWalFailoverSpec(ctx context.Context, clientset kubernetes.Interface, sts *appsv1.StatefulSet, nodeName string, nodeIdx int32, input *parsedMigrationInput) {
```
NishanthNalluri: This can be a private method.

build/templates/cockroachdb-parent/charts/operator/values.yaml
shreyaskm623: Will fix this now as the other PR is merged.

pkg/migrate/build-manifests.go
```diff
@@ -183,6 +186,11 @@ func (m *Manifest) FromHelmChart() error {
 			return errors.Newf("pod %s isn't scheduled to a node", podName)
 		}
 
+		// Build the walFailoverSpec if applicable
+		buildWalFailoverSpec(ctx, m.clientset, sts, pod.Spec.NodeName, nodeIdx, &input)
+		if input.walFailoverSpec != nil && clusterWalSpec == nil {
```
prafull01: Why do we need clusterWalSpec == nil condition?

pkg/migrate/helpers_test.go
```diff
@@ -368,3 +372,562 @@ func TestDetectPCRFromInitJob(t *testing.T) {
...
+			},
+		},
+	}
+
+	testCases := []struct {
+		name           string
+		walFailoverArg string
+		expectedPath   string
+		expectedStatus string
+		expectWarning  bool
```
prafull01: You are not using expectWarning anywhere. Remove it.

pkg/upstream/cockroach-operator/api/v1alpha1/crdbcluster_types_test.go
```diff
@@ -0,0 +1,261 @@
+package v1alpha1
```
prafull01: No need to keep this entire test file. The tests should reside in the cockroach-enterprise-operator repository.
The types.go files are included here temporarily to provide struct access for easier code development, as that repository isn’t public yet.

These files will be removed once the repository becomes public.

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
```diff
@@ -173,6 +173,11 @@ func (o *PublicOperatorToCockroachEnterpriseOperator) TestDefaultMigration(t *te
 	// Verify priority class migration
 	verifyOperatorMigration(t, kubectlOptions, o.CustomResourceBuilder.Cr())
 
+	// Wait for ingress controller to sync the updated ingress configuration
+	// after Helm install modified the ingress resources
+	t.Log("Waiting for ingress controller to sync after migration")
+	time.Sleep(20 * time.Second)
```
prafull01: Do we need this? It is working fine now. What has changed in this PR for this sleep to be introduced?

pkg/migrate/helpers.go
```diff
@@ -369,13 +376,105 @@ func detectPCRFromInitJob(clientset kubernetes.Interface, stsName, namespace str
...
+			if !strings.HasPrefix(f, "--wal-failover=") {
+				continue
+			}
+
+			val := strings.TrimPrefix(f, "--wal-failover=")
+			if val == "disabled" {
+				// do not include the walFailoverSpec - when the wal-failover is disabled
+				input.walFailoverSpec = nil
+			} else if val == "among-stores" {
+				fmt.Printf("Multi-store configuration (among-stores) is not supported in the new operator\n")
```
prafull01: Lets panic from here, as we don't support migration of multi store so the migration helper stop right here.
`panic("💥 PANIC: Multi-store configuration (among-stores) is not supported in the new operator")`

tests/e2e/migrate/public_operator_to_cockroach_enterprise_operator_test.go
NishanthNalluri: @shreyaskm623  This is not needed as the ingress fix is related to podLabels.

PR #567, opened 2025-10-20
[CRDB-54713] Add support for init-containers and custom volumes in migration

prafull01:


pkg/migrate/helpers.go
```diff
@@ -580,6 +587,23 @@ func generateParsedMigrationInput(
 		}
 	}
 
+	ignoreInitContainers := []string{"copy-certs"}
```
pritesh-lahoti: nit: could these lists be global variables, so that it's easy to reference and update in the future?

PR #571, opened 2025-10-28
Fixes the slack notification being sent for custom branches

shreyaskm623:


.github/workflows/integration-tests.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
             test_output.log
             test_output.json
       - name: Slack notification (multi-region)
-        if: ${{ always() }}
+        if: ${{ always() && github.ref_name == 'master' }}
```
NishanthNalluri, line 52: I am not sure of what always() means here but cant we just have `github.ref_name == 'master'` as the if condition?

shreyaskm623: always() ensures that the step runs irrespective of the previous test failure.
In our case it's archive test result , I believe we need to have it. Otherwise is the archive test result fails , it's possible that slack notification might not be sent

PR #577, opened 2025-11-20
[CRDB-53973] e2e/operator: advanced e2e tests for bare-metal infrastructure(kind).

NishanthNalluri:
This commit introduces advanced e2e tests for single and multi region for CockroachDB features like Wal-failover, EAR and PCR. These tests verifies supportibility when configured through helm-charts rather than re-validating functionality.

Implements [CRDB-53973](https://cockroachlabs.atlassian.net/browse/CRDB-53973)

### DIFF
--- a/tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
+++ b/tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
@@ -10,6 +10,7 @@ import (
 
 	api "github.com/cockroachdb/cockroach-operator/apis/v1alpha1"
 	"github.com/cockroachdb/helm-charts/tests/e2e/operator"
+	"github.com/cockroachdb/helm-charts/tests/e2e/operator/infra"
 	"github.com/cockroachdb/helm-charts/tests/testutil"
 	"github.com/cockroachdb/helm-charts/tests/testutil/migration"
 	"github.com/gruntwork-io/terratest/modules/helm"
@@ -56,6 +57,21 @@ func init() {
 	})
 	if err != nil {
 		panic(err)
+	}
+}
+
+// providerCloudRegion returns the cloud region for the active provider.
+func providerCloudRegion() string {
+	p := strings.TrimSpace(strings.ToLower(os.Getenv("PROVIDER")))
+	switch p {
+	case "k3d", "": // default to k3d when PROVIDER is unset
+		return infra.RegionCodes[infra.ProviderK3D][0]
+	case "kind":
+		return infra.RegionCodes[infra.ProviderKind][0]
+	case "gcp":
+		return infra.RegionCodes[infra.ProviderGCP][0]
+	default:
+		return ""
 	}
 }
 
@@ -145,7 +161,7 @@ func (h *HelmChartToOperator) TestDefaultMigration(t *testing.T) {
 		k8s.RunKubectl(t, kubectlOptions, "delete", "priorityclass", "crdb-critical")
 	}()
 
-	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions, providerCloudRegion())
 	defer func() {
 		t.Log("Uninstall the cockroachdb enterprise operator")
 		operator.UninstallCockroachDBEnterpriseOperator(t, kubectlOptions)
@@ -263,7 +279,7 @@ func (h *HelmChartToOperator) TestCertManagerMigration(t *testing.T) {
 		k8s.RunKubectl(t, kubectlOptions, "delete", "priorityclass", "crdb-critical")
 	}()
 
-	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions, providerCloudRegion())
 	defer func() {
 		t.Log("Uninstall the cockroachdb enterprise operator")
 		operator.UninstallCockroachDBEnterpriseOperator(t, kubectlOptions)
@@ -381,7 +397,7 @@ func (h *HelmChartToOperator) TestDedicatedLogsPVCMigration(t *testing.T) {
 		k8s.RunKubectl(t, kubectlOptions, "delete", "priorityclass", "crdb-critical")
 	}()
 
-	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions, providerCloudRegion())
 	defer func() {
 		t.Log("Uninstall the cockroachdb enterprise operator")
 		operator.UninstallCockroachDBEnterpriseOperator(t, kubectlOptions)
@@ -476,7 +492,7 @@ func (h *HelmChartToOperator) TestPCRPrimaryMigration(t *testing.T) {
 		k8s.RunKubectl(t, kubectlOptions, "delete", "priorityclass", "crdb-critical")
 	}()
 
-	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions)
+	operator.InstallCockroachDBEnterpriseOperator(t, kubectlOptions, providerCloudRegion())
 	defer func() {
 		t.Log("Uninstall the cockroachdb enterprise operator")
 		operator.UninstallCockroachDBEnterpriseOperator(t, kubectlOptions)
--- a/tests/e2e/operator/multiRegion/cockroachdb_multi_region_advanced_features_test.go
+++ b/tests/e2e/operator/multiRegion/cockroachdb_multi_region_advanced_features_test.go
@@ -1,0 +1,250 @@
+package multiRegion
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/cockroachdb/helm-charts/tests/e2e/operator"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// TestWALFailoverMultiRegion tests WAL failover with different paths in each region
+// Region 0: WAL failover enabled with custom path
+// Region 1: WAL failover disabled
+func (r *multiRegion) TestWALFailoverMultiRegion(t *testing.T) {
+	// Setup namespaces and CA for each region
+	cleanup := r.SetupMultiClusterWithCA(t)
+	defer cleanup()
+
+	// Region 0: Install with WAL failover enabled
+	cluster0 := r.Clusters[0]
+	walPath0 := "/cockroach/wal-region-0"
+
+	t.Logf("Installing region 0 (%s) with WAL failover enabled at path %s", cluster0, walPath0)
+	config0 := operator.AdvancedInstallConfig{
+		WALFailoverEnabled: true,
+		WALFailoverSize:    "5Gi",
+		CustomValues: map[string]string{
+			"cockroachdb.crdbCluster.walFailoverSpec.path": walPath0,
+		},
+	}
+	r.InstallChartsWithAdvancedConfig(t, cluster0, 0, config0)
+
+	// Region 1: Install without WAL failover
+	cluster1 := r.Clusters[1]
+	t.Logf("Installing region 1 (%s) without WAL failover", cluster1)
+	config1 := operator.AdvancedInstallConfig{}
+	r.InstallChartsWithAdvancedConfig(t, cluster1, 1, config1)
+
+	// Validate CockroachDB cluster health in both regions
+	for _, cluster := range r.Clusters {
+		r.ValidateCRDB(t, cluster)
+	}
+
+	// Validate multi-region setup
+	r.ValidateMultiRegionSetup(t)
+
+	// Validate WAL failover in region 0
+	t.Log("Validating WAL failover in region 0")
+	r.ValidateWALFailover(t, cluster0, &operator.AdvancedValidationConfig{
+		WALFailover: operator.WALFailoverValidation{
+			CustomPath: walPath0,
+		},
+	})
+
+	// Validate NO WAL failover in region 1
+	t.Log("Validating NO WAL failover in region 1")
+	kubeConfig, _ := r.GetCurrentContext(t)
+	kubectlOptions1 := k8s.NewKubectlOptions(cluster1, kubeConfig, r.Namespace[cluster1])
+
+	pods := k8s.ListPods(t, kubectlOptions1, metav1.ListOptions{
+		LabelSelector: operator.LabelSelector,
+	})
+	require.True(t, len(pods) > 0, "No CockroachDB pods found in region 1")
+
+	podCommand, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions1,
+		"get", "pod", pods[0].Name, "-o", "jsonpath={.spec.containers[?(@.name=='cockroachdb')].command}")
+	require.NoError(t, err)
+	require.NotContains(t, podCommand, "--wal-failover", "Region 1 should not have WAL failover enabled")
+	t.Log("Confirmed region 1 does not have WAL failover")
+
+	t.Logf("WAL failover multi-region test completed successfully")
+}
+
+// TestEncryptionAtRestMultiRegion tests encryption at rest with different secrets per region
+// Region 0: Encryption enabled with secret "cmek-key-secret-region-0"
+// Region 1: Encryption disabled (no encryption)
+func (r *multiRegion) TestEncryptionAtRestMultiRegion(t *testing.T) {
+	// Setup namespaces and CA for each region
+	cleanup := r.SetupMultiClusterWithCA(t)
+	defer cleanup()
+
+	// Generate encryption key for region 0
+	encryptionKeyB64 := r.GenerateEncryptionKey(t)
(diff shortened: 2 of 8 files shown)


tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	t.Log("Creating admin user on standby cluster...")
+	createAdminSQL := fmt.Sprintf("CREATE USER IF NOT EXISTS %s WITH PASSWORD '%s'", "pcr_admin", "admin_password_123")
+	_, err = execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", createAdminSQL)
+	require.NoError(t, err)
+	_, err = execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", "GRANT admin TO pcr_admin")
+	require.NoError(t, err)
+	t.Log("✓ Created user pcr_admin with admin privileges")
+	t.Log("Standby cluster setup complete!")
+
+	// Step 4: Set up replication stream from standby to primary
```
nameisbhaskar:
```suggestion
	// Step 3: Set up replication stream from standby to primary
```

nameisbhaskar (approved):
LGTM! One minor nit comment.

NishanthNalluri:
We need to add a new github workflow for running these advanced tests right?

tests/e2e/operator/region.go
```diff
@@ -445,10 +449,32 @@ func (r *Region) CleanupResources(t *testing.T) {
...
+		// conflicts in subsequent runs, even when helm delete fails or is incomplete.
+		nodeReaderName := fmt.Sprintf("cockroachdb-%s-node-reader", namespace)
+		clusterScopedResources := [][]string{
+			{"priorityclass", "cockroach-operator"},
+			{"priorityclass", "cockroachdb"},
+			{"clusterrole", "cockroach-operator-role"},
+			{"clusterrolebinding", "cockroach-operator-default"},
+			{"clusterrole", nodeReaderName},
+			{"clusterrolebinding", nodeReaderName},
+			{"mutatingwebhookconfiguration", "cockroach-operator-mutating-webhook-configuration"},
```
NishanthNalluri: With introduction of namespace scoping, the names would differ for these webhooks, we can remove these as uninstalling operator will clean these resources.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	return fmt.Sprintf("postgresql://root:root@localhost:26257?options=-ccluster%%3D%s", cluster)
+}
+
+// generateExternalConnectionURI creates external connection URI using cockroach encode-uri
+// This is used for cross-cluster connections (e.g., replication from primary to standby)
+// Format: cockroach encode-uri 'postgresql://USERNAME:PASSWORD@HOST' [flags]
+func generateExternalConnectionURI(t *testing.T, kubectlOpts *k8s.KubectlOptions, pod string, connString string) string {
+	output, err := k8s.RunKubectlAndGetOutputE(t, kubectlOpts,
+		"exec", pod, "-c", "cockroachdb", "--",
+		"/cockroach/cockroach", "encode-uri",
```
NishanthNalluri: `encode-uri` is deprecated from 26.2, use `convert-url` instead.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	t.Log("==================================================")
+	t.Log("Step 7: Testing failover (promoting standby to primary)")
+	t.Log("==================================================")
+
+	// Stop the service first (if running in SHARED mode)
+	t.Log("Stopping main virtual cluster service on standby before cutover...")
+	_, _ = execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", "ALTER VIRTUAL CLUSTER main STOP SERVICE")
+	t.Log("✓ Service stopped")
+
+	time.Sleep(5 * time.Second)
```
NishanthNalluri: Lets try to use retry.DoWithRetryE pattern everywhere and eliminate time.Sleep unless there is no other way.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	t.Log("Completing replication to latest (this promotes standby to be writable)...")
+	_, err = execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", "ALTER VIRTUAL CLUSTER main COMPLETE REPLICATION TO LATEST")
+	if err != nil {
+		t.Logf("Note: Cutover command returned: %v", err)
+	}
+	t.Log("✓ Replication completed to latest")
+
+	// Wait for cutover to complete
+	t.Log("Waiting for cutover to complete (15 seconds)...")
+	time.Sleep(15 * time.Second)
```
NishanthNalluri: Same as above

tests/e2e/operator/region.go
NishanthNalluri: And even other cluster resources also gets provisioned with different names with namespace scoping

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	t.Log("Enabling rangefeed on primary cluster (required for PCR)...")
+	_, err := execSQLOnVC(t, primaryKubectlOptions, primaryPod, primarySystemURI, "", "SET CLUSTER SETTING kv.rangefeed.enabled = true")
+	require.NoError(t, err)
+	t.Log("✓ Rangefeed enabled on primary cluster")
+
+	// With --virtualized flag, the 'main' virtual cluster is created automatically
+	// We just need to verify it exists and ensure it's in SHARED mode
+	t.Log("Verifying main virtual cluster exists on primary (automatically created with --virtualized flag)...")
+	primaryVCs, err := execSQLOnVC(t, primaryKubectlOptions, primaryPod, primarySystemURI, "", "SHOW VIRTUAL CLUSTERS")
+	require.NoError(t, err)
```
NishanthNalluri, line 1232: Did we add validations to verify the presence of virtualised flags in the start command?

tests/e2e/operator/region.go
NishanthNalluri: Do we need these two here? I think we are using InstallChartsWithAdvancedConfig for PCR tests.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	readerVCsStatus, err := execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", "SHOW VIRTUAL CLUSTERS")
+	require.NoError(t, err)
+	require.Contains(t, readerVCsStatus, "main-reader", "Standby should have main-reader virtual cluster")
+	t.Logf("Virtual clusters on standby:\n%s", readerVCsStatus)
+	t.Log("✓ Reader virtual cluster is ready")
+
+	// Step 6d: Read from the reader virtual cluster
+	t.Log("Attempting to read replicated data from reader virtual cluster...")
+	readerURI := generateLocalTenantURI("main-reader")
+	readFromStandby, err := execSQLOnVC(t, standbyKubectlOptions, standbyPod, readerURI, "bank", "SELECT id, balance FROM accounts ORDER BY id")
```
NishanthNalluri: Looks like we are testing the actual PCR functionality, if so are we sure all these validations are correct? Are we creating the bank db and accounts table in main tenant?

NishanthNalluri: I believe its testdb?

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	t.Log("Creating external connection on standby cluster...")
+	createExternalConnSQL := fmt.Sprintf("CREATE EXTERNAL CONNECTION IF NOT EXISTS primary_replication AS '%s'", encodedPrimaryURI)
+	_, err = execSQLOnVC(t, standbyKubectlOptions, standbyPod, standbySystemURI, "", createExternalConnSQL)
+	require.NoError(t, err)
+	t.Log("✓ External connection 'primary_replication' created on standby")
+
+	// Step 4c: Create the replication stream using the external connection name
+	t.Log("Creating replication stream on standby cluster using external connection...")
+	t.Log("This will create the main virtual cluster on standby and start replicating data from primary")
+	createReplicationCmd := "CREATE VIRTUAL CLUSTER main FROM REPLICATION OF main ON 'external://primary_replication'"
```
NishanthNalluri: Can we declare all these commands at once place, will be easier to understand and can refer via variables.

tests/e2e/operator/region.go
```diff
@@ -592,3 +623,856 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+		"Pod command should contain %s", expectedFlag)
+	t.Logf("Cockroach start command contains --wal-failover flag with path %s", expectedPath)
+
+	// 2. Verify COCKROACH_WAL_FAILOVER environment variable is set
+	t.Log("Verifying COCKROACH_WAL_FAILOVER environment variable...")
+	walFailoverEnv, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
+		"get", "pod", podName, "-o", "jsonpath={.spec.containers[?(@.name=='cockroachdb')].env[?(@.name=='COCKROACH_WAL_FAILOVER')].value}")
+	if err == nil && walFailoverEnv != "" {
+		t.Logf("COCKROACH_WAL_FAILOVER environment variable is set to: %s", walFailoverEnv)
+	}
```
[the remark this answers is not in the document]
NishanthNalluri: Is this not a mandatory env thats needs to be present when wal-failover flag is set?

tests/e2e/operator/singleRegion/cockroachdb_single_region_advanced_features_test.go
```diff
@@ -0,0 +1,590 @@
...
+	r.ValidateEncryptionAtRest(t, cluster, &operator.AdvancedValidationConfig{
+		EncryptionAtRest: operator.EncryptionAtRestValidation{
+			SecretName:     "cmek-key-secret-new",
+			OldKeyExpected: "old-key=/etc/cockroach-key/",
+		},
+	})
+
+	// Verify new secret exists and is being used
+	newSecretData, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
+		"get", "secret", "cmek-key-secret-new", "-o", "jsonpath={.data.StoreKeyData}")
```
NishanthNalluri: I see this and `cmek-key-secret` hardcoded many times, can we make it a const?

tests/e2e/operator/multiRegion/cockroachdb_multi_region_advanced_features_test.go
```diff
@@ -0,0 +1,238 @@
...
+	t.Log("Installing primary multi-region cluster...")
+	for i, cluster := range r.Clusters {
+		primaryConfig := operator.AdvancedInstallConfig{
+			VirtualClusterMode: "primary",
+		}
+		r.InstallChartsWithAdvancedConfig(t, cluster, i, primaryConfig)
+	}
+
+	// Validate primary cluster health in all regions
+	r.VirtualClusterModePrimary = true
```
NishanthNalluri: Are we using this in `ValidateCRDB`, i can see we are only checking `VirtualClusterModeStandby`

tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
```diff
@@ -59,6 +60,21 @@ func init() {
 	}
 }
 
+// providerCloudRegion returns the cloud region for the active provider.
+func providerCloudRegion() string {
```
NishanthNalluri, line 64: Worth moving this to shared package, fine if we are achieving these refactoring's in upcoming PR's.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+
+	// 5. Verify encryption algorithm via node metrics.
+	// rocksdb.encryption.algorithm: 0=Plaintext, 1=AES-128-CTR, 2=AES-192-CTR, 3=AES-256-CTR
+	t.Log("Verifying encryption algorithm via node metrics...")
+	encAlgorithm, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
+		"exec", podName, "-c", "cockroachdb", "--",
+		"/cockroach/cockroach", "sql",
+		"--certs-dir=/cockroach/cockroach-certs",
+		"--host=localhost:26257",
+		"-e", "SET allow_unsafe_internals = true; SELECT value FROM crdb_internal.node_metrics WHERE name = 'rocksdb.encryption.algorithm';")
```
NishanthNalluri: Have we thought of this validation


```
 out, err := k8s.RunKubectlAndGetOutputE(t, kubectlOpts,
      "exec", podName, "-c", "cockroachdb", "--",
      "/cockroach/cockroach", "debug", "encryption-active-key", "/cockroach/cockroach-data")
```

NishanthNalluri: Do we have similar metric with pebble? something like `pebble.encryption.algorithm`.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+	require.Contains(t, podSpec, "key=/etc/cockroach-key/", "Encryption flag should contain key path")
+	oldKeyExpected := validationConfig.EncryptionAtRest.OldKeyExpected
+	require.Contains(t, podSpec, oldKeyExpected, "Encryption flag should contain %s", oldKeyExpected)
+	t.Log("Encryption flag verified: path, key, and old-key are all present")
+
+	// 5. Verify encryption algorithm via node metrics.
+	// rocksdb.encryption.algorithm: 0=Plaintext, 1=AES-128-CTR, 2=AES-192-CTR, 3=AES-256-CTR
+	t.Log("Verifying encryption algorithm via node metrics...")
+	encAlgorithm, err := k8s.RunKubectlAndGetOutputE(t, kubectlOptions,
+		"exec", podName, "-c", "cockroachdb", "--",
```
NishanthNalluri: Just like execSQLOnVC, lets have a common method for running sql, cockroach and exec commands in cockroachdb container.

tests/e2e/operator/region.go
```diff
@@ -600,3 +631,814 @@ func PatchHelmValues(inputValues map[string]string) map[string]string {
...
+		"/cockroach/cockroach", "sql",
+		"--certs-dir=/cockroach/cockroach-certs",
+		"--host=localhost:26257",
+		"-e", "SET allow_unsafe_internals = true; SELECT value FROM crdb_internal.node_metrics WHERE name = 'rocksdb.encryption.algorithm';")
+	require.NoError(t, err)
+	require.Contains(t, encAlgorithm, "3", "rocksdb.encryption.algorithm should be 3 (AES-256-CTR)")
```
NishanthNalluri, line 1114: SInce the output is a virtual table can we extract as a json and check `encryption_type`? If it emits in the table/

tests/e2e/operator/region.go
shreyaskm623: Yes, I have tried this. I'm using the v26+ version of the cockroachDB, by running the above command it should ideally return something like below when the EAR is enabled - 
```
AES256_CTR:a99571d0ae8da822734a178368af6c0bd54210dac3f4e1059f675db61c32ecbf
```

But due to some issue the command doesn't work correctly and returns - 
```
Plaintext:
```

So, we cannot use it.

shreyaskm623: No, I couldn't find, I tried running the below query on the cluster - 
```
SELECT name, value FROM crdb_internal.node_metrics                                                                                                                                         
WHERE name LIKE 'pebble.%'                        
ORDER BY name;
```
But it did not return any results here.

tests/e2e/operator/region.go
shreyaskm623: `crdb_internal.node_metrics` only has 3 columns — store_id, name, value. There is no `encryption_type` field in this table or anywhere in `crdb_internal`. The encryption algorithm is only emitted as a numeric FLOAT8 value 3 (AES-256-CTR).

tests/e2e/migrate/helm_chart_to_cockroach_enterprise_operator_test.go
shreyaskm623: Will be taking it as part of the GCP PR.

tests/e2e/operator/region.go
shreyaskm623: The flags - `--virtualized` and `--virtualized-empty` are passed by the operator as part of the `cockroach init` command, unlike in the case of WalFailover we store the configuration in the CR spec, we don't do it for PCR. That's why I'd relied on other set of validations. However, in the `TestHelmInstallVirtualCluster` the flags are validated by checking the container logs - 
```
"/cockroach/cockroach init --host :26258 --virtualized"
```
I can add this if at all we need the flag validations.

tests/e2e/operator/region.go
shreyaskm623: I have addressed this, but works fine only for the v26.2+ images. For the current v26.1.4 image also it fails.
```
cockroach convert-url in v26.1.4 doesn't support --ca-cert and --inline flags — those flags exist on encode-uri.
```

NishanthNalluri (commented):
This looks nice now! Thanks for addressing all the reviews.

PR #608, opened 2026-03-13
[CRDB-55385] self-signer: Add support for additional SANs in certificates

nameisbhaskar:
CRDB-55385: Allow users to specify additional Subject Alternative Names (SANs) for node certificates generated by the self-signer utility.

This enables customers to:
- Route traffic through load balancers with sslmode=verify-full
- Support Physical Cluster Replication (PCR) via load balancers
- Include external endpoints in certificate validation

Implementation:
- Add additionalSANs field to Helm values (tls.certs.selfSigner.additionalSANs)
- Pass ADDITIONAL_SANS environment variable to self-signer jobs
- Parse and append additional SANs to node certificate hosts list
- Update schema validation for additionalSANs array
- Apply to both certificate generation and rotation

Example usage:
  tls: certs: selfSigner: additionalSANs: - "my-loadbalancer.example.com" - "192.168.1.100"

### DIFF
--- a/build/templates/cockroachdb/values.yaml
+++ b/build/templates/cockroachdb/values.yaml
@@ -633,6 +633,12 @@ tls:
       nodeCertDuration: 8760h
       # Expiry window of node certificates means a window before actual expiry in which node certs should be rotated.
       nodeCertExpiryWindow: 168h
+      # Additional Subject Alternative Names (SANs) to include in node certificates.
+      # Useful for load balancers or external endpoints that need to be included in the certificate.
+      # Example: ["my-loadbalancer.example.com", "10.20.30.40"]
+      # Note: Changes to this value will only take effect after the next certificate rotation cycle,
+      # or when certificates are manually regenerated (e.g., by deleting the node-secret).
+      additionalSANs: []
       # If set, the cockroachdb cert selfSigner will rotate the certificates before expiry.
       rotateCerts: true
       # Wait time for each cockroachdb replica to become ready once it comes in running state. Only considered when rotateCerts is set to true
@@ -688,7 +694,7 @@ tls:
     # Image Placeholder for the selfSigner utility. This will be changed once the CI workflows for the image is in place.
     image:
       repository: cockroachlabs-helm-charts/cockroach-self-signer-cert
-      tag: "1.9"
+      tag: "1.10"
       pullPolicy: IfNotPresent
       registry: gcr.io
       credentials: {}
--- a/cmd/self-signer/root.go
+++ b/cmd/self-signer/root.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -82,8 +83,12 @@ func init() {
 	}
 }
 
-func getInitialConfig(caDuration, caExpiry, nodeDuration, nodeExpiry, clientDuration,
-	clientExpiry string) (generator.GenerateCert, error) {
+// getInitialConfig initializes a GenerateCert instance from certificate durations
+// and environment variables (STATEFULSET_NAME, CLUSTER_DOMAIN, ADDITIONAL_SANS).
+func getInitialConfig(
+	caDuration, caExpiry, nodeDuration, nodeExpiry, clientDuration,
+	clientExpiry string,
+) (generator.GenerateCert, error) {
 
 	genCert := generator.NewGenerateCert(cl)
 
@@ -113,6 +118,13 @@ func getInitialConfig(caDuration, caExpiry, nodeDuration, nodeExpiry, clientDura
 			return genCert, errors.New("Required CLUSTER_DOMAIN env not found")
 		}
 		genCert.ClusterDomain = domain
+
+		// Parse additional SANs if provided
+		if additionalSANsStr, exists := os.LookupEnv("ADDITIONAL_SANS"); exists && additionalSANsStr != "" {
+			sans := strings.Split(additionalSANsStr, ",")
+			// Use shared sanitization helper to trim whitespace and filter out empty strings
+			genCert.AdditionalSANs = generator.SanitizeAdditionalSANs(sans)
+		}
 	}
 
 	return genCert, nil
--- a/cockroachdb-parent/charts/cockroachdb/values.yaml
+++ b/cockroachdb-parent/charts/cockroachdb/values.yaml
@@ -49,6 +49,12 @@ cockroachdb:
       # podUpdateTimeout specifies the timeout for pods to reach the running state.
       # Only considered when rotateCerts is true.
       podUpdateTimeout: 2m
+      # additionalSANs specifies additional Subject Alternative Names (SANs) to include in node certificates.
+      # This is useful for load balancers or external endpoints that need SSL verification.
+      # Example: ["my-loadbalancer.example.com", "10.20.30.40"]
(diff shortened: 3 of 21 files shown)


pkg/generator/generate_cert.go
```diff
@@ -548,6 +570,13 @@ func (rc *GenerateCert) GenerateNodeCert(ctx context.Context, nodeSecretName, na
 		hosts = append(hosts, operatorJoinServiceHosts...)
 	}
 
+	// Append additional user-provided SANs (after sanitizing)
+	validSANs := SanitizeAdditionalSANs(rc.AdditionalSANs)
```
NishanthNalluri, line 574: Do we need to call this again, since its already called at the root level?

pkg/generator/generate_cert_test.go
```diff
@@ -0,0 +1,426 @@
...
+		certsDir, cleanup := tempDir(t)
+		defer cleanup()
+		caKey := filepath.Join(certsDir, "ca.key")
+
+		// Create CA first
+		err := security.CreateCAPair(certsDir, caKey, defaultKeySize, testCALifetime, true, true)
+		require.NoError(t, err, "Failed to create CA pair")
+
+		// Test with SANs that have whitespace, empty strings, and valid entries
+		genCert := GenerateCert{
```
NishanthNalluri, line 273: Can we move the duplicated configs, code paths to a helper? And if feasible, we can convert these tests into table-driven format.

pkg/generator/generate_cert_test.go
```diff
@@ -0,0 +1,426 @@
...
+		for _, san := range standardSANs {
+			assert.Contains(t, cert.DNSNames, san, "Standard SAN %s should be present", san)
+		}
+
+		// Verify no unexpected additional SANs
+		assert.NotContains(t, cert.DNSNames, "my-loadbalancer.example.com", "Should not contain additional SANs")
+		assert.NotContains(t, cert.DNSNames, "backup-lb.example.com", "Should not contain additional SANs")
+	})
+
+	t.Run("EmptyAdditionalSANs", func(t *testing.T) {
```
NishanthNalluri, line 225: This just tests whether the hosts are valid SAN's or not right? i could see the similar test case being covered in `TestSanitizeAdditionalSANs`.

cmd/self-signer/root.go
```diff
@@ -82,8 +83,24 @@ func init() {
...
+// durations, service names, and optional additional SANs from environment variables.
+//
+// It reads the following environment variables:
+//   - STATEFULSET_NAME: Name of the StatefulSet (required for non-client-only mode)
+//   - CLUSTER_DOMAIN: Kubernetes cluster domain (required for non-client-only mode)
+//   - ADDITIONAL_SANS: Comma-separated list of additional SANs for node certificates (optional)
+//
+// The function configures certificate durations for CA, node, and client certificates,
+// derives service names from the StatefulSet name, and parses any additional SANs
+// to be included in node certificates.
```
NishanthNalluri, line 96: Do we actually need explaining these env vars? Looks like they are self-explanatory and also covered in the values.yaml/README

tests/testutil/migration/helm.go
```diff
@@ -60,6 +60,22 @@ func (h *HelmInstall) InstallHelm(t *testing.T) {
...
+	hasTimeout := false
+	for _, arg := range h.HelmOptions.ExtraArgs["install"] {
+		if arg == "--timeout" || arg == "-t" || strings.HasPrefix(arg, "--timeout=") {
+			hasTimeout = true
+			break
+		}
+	}
+	if !hasTimeout {
+		h.HelmOptions.ExtraArgs["install"] = append(h.HelmOptions.ExtraArgs["install"], "--timeout", "10m")
+	}
```
NishanthNalluri, line 77: Since every test can have additional timeout through extraArgs, i believe we dont need these changes anymore?

cockroachdb/values.yaml
```diff
@@ -625,6 +625,10 @@ tls:
       nodeCertDuration: 8760h
       # Expiry window of node certificates means a window before actual expiry in which node certs should be rotated.
       nodeCertExpiryWindow: 168h
+      # Additional Subject Alternative Names (SANs) to include in node certificates.
+      # Useful for load balancers or external endpoints that need to be included in the certificate.
+      # Example: ["my-loadbalancer.example.com", "10.20.30.40"]
+      additionalSANs: []
```
NishanthNalluri, line 631: Does the certs get updated with additional san's immediately or wait for the next rotation cycle? I guess its latter as we wont generate new certs until rotation is due. So do we need to update that logic?

cockroachdb/templates/job-certSelfSigner.yaml
```diff
@@ -80,6 +80,10 @@ spec:
             value: {{ .Release.Namespace | quote }}
           - name: CLUSTER_DOMAIN
             value: {{ .Values.clusterDomain}}
+          {{ if .Values.tls.certs.selfSigner.additionalSANs }}
+          - name: ADDITIONAL_SANS
+            value: {{ .Values.tls.certs.selfSigner.additionalSANs | join "," | quote }}
+          {{ end }}
```
NishanthNalluri, line 86: Nit: Use {{- for trimming whitespaces.

pkg/generator/generate_cert.go
nameisbhaskar: You are right. Thanks for pointing out.

pkg/generator/generate_cert_test.go
nameisbhaskar: I will remove the test.

cockroachdb/values.yaml
nameisbhaskar: You're correct - certificates won't be updated immediately when `additionalSANs` is changed. The new SANs will only take effect after the next rotation cycle (when certificates are approaching expiry) or if certificates are manually regenerated.
I will add a comment to the values.yaml file to set proper user expectations.

nameisbhaskar:
Thanks @NishanthNalluri for the review. I have addressed the comments. PTAL!

build/templates/cockroachdb/values.yaml
```diff
@@ -624,6 +624,12 @@ tls:
       nodeCertDuration: 8760h
       # Expiry window of node certificates means a window before actual expiry in which node certs should be rotated.
       nodeCertExpiryWindow: 168h
+      # Additional Subject Alternative Names (SANs) to include in node certificates.
+      # Useful for load balancers or external endpoints that need to be included in the certificate.
+      # Example: ["my-loadbalancer.example.com", "10.20.30.40"]
+      # Note: Changes to this value will only take effect after the next certificate rotation cycle,
+      # or when certificates are manually regenerated (e.g., by deleting the node-secret).
+      additionalSANs: []
       # If set, the cockroachdb cert selfSigner will rotate the certificates before expiry.
```
NishanthNalluri, line 642: We can add this capability to the new crdb charts as well right?

cockroachdb-parent/charts/cockroachdb/values.yaml
```diff
@@ -69,7 +69,7 @@ cockroachdb:
         # repository defines the image repository for the self-signer container.
         repository: cockroachlabs-helm-charts/cockroach-self-signer-cert
         # tag defines the image tag for the self-signer container.
-        tag: "1.9"
+        tag: "1.11"
```
NishanthNalluri: Do we need to revert this to 1.9?

NishanthNalluri (approved):
Thanks for addressing reviews!!

PR #624, opened 2026-04-25
[CRDB-53890] e2e: add OpenShift infrastructure support for single and multi-region operator tests

shreyaskm623:
Extends the e2e test framework with a new `openshift` provider that provisions OpenShift clusters on GCP via `openshift-install` and runs the existing single-region and multi-region operator test suites against them.

JIRA: https://cockroachlabs.atlassian.net/browse/CC-35910

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,8 @@ bin/
 build/artifacts
 bazel-*
 /vendor/
+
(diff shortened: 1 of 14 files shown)


shreyaskm623:
All the Copilot reviews have been addressed here already - https://github.com/cockroachdb/helm-charts/pull/607

PR #634, opened 2026-05-16
ci: declare workflow-level `contents: read` on 3 workflows

arpitjain099:
Pins the default `GITHUB_TOKEN` to `contents: read` on 3 workflows in `.github/workflows/` that don't call a GitHub API beyond the initial checkout.

## Why

CVE-2025-30066 (March 2025 `tj-actions/changed-files` supply-chain compromise) exfiltrated `GITHUB_TOKEN` from workflow logs. Pinning per workflow caps runtime authority irrespective of the repo or org default, gives drift protection if the default ever widens, and is credited per-file by the OpenSSF Scorecard `Token-Permissions` check.

YAML validated locally with `yaml.safe_load` on each touched file.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,6 +7,9 @@ on:
     tags:
       - 'v*.*.*'
 
+permissions:
+  contents: read
+
 jobs:
 
   needs-build:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,9 @@ on:
     branches:
       - '*'
 
+permissions:
+  contents: read
+
 jobs:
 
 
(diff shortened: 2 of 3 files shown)


NishanthNalluri (approved):

NishanthNalluri:
@arpitjain099 please update job-level permission on detect-self-signer-change in `ci.yaml` with 

```  
detect-self-signer-change:
    permissions:
      contents: read
      pull-requests: read

arpitjain099:
Thanks @NishanthNalluri, pushed the per-job `permissions: contents: read + pull-requests: read` on `detect-self-signer-change`. Made sense because of the `dorny/paths-filter` step looking at PR-changed files. Let me know if anything else.

PR #635, opened 2026-05-17
e2e/operator: Fix GCP single-region and multi-region test infrastructure

shreyaskm623:
Fixes the existing e2e tests running in GCP for the single and multi region

### DIFF
--- a/tests/e2e/operator/infra/gcp.go
+++ b/tests/e2e/operator/infra/gcp.go
@@ -44,6 +44,12 @@ const (
 // Default project ID to use if not specified in the environment.
 const defaultProjectID = "helm-testing"
 
+// getNodeServiceAccount returns the GKE node service account from the environment.
+// If empty, gcloud falls back to the project's default Compute Engine service account.
+func getNodeServiceAccount() string {
+	return os.Getenv("GCP_NODE_SERVICE_ACCOUNT")
+}
+
 // getProjectID returns the GCP project ID from the environment variable or falls back to default.
 func getProjectID() string {
 	if projectID := os.Getenv("GCP_PROJECT_ID"); projectID != "" {
@@ -635,6 +641,10 @@ func createGKERegionalCluster(ctx context.Context, client *container.Service, se
 		"--machine-type", gcpDefaultMachineType,
 		"--disk-size", "30GB", // Limit disk size to 30GB
 		"--quiet", // Suppress interactive prompts
+	}
+
+	if sa := getNodeServiceAccount(); sa != "" {
+		args = append(args, "--service-account", sa)
 	}
 
 	cmd := exec.Command("gcloud", args...)
--- a/tests/e2e/operator/region.go
+++ b/tests/e2e/operator/region.go
@@ -365,6 +365,8 @@ func (r *Region) ValidateCRDBContainerResources(t *testing.T, kubectlOptions *k8
 
 // CreateCACertificate creates CA cert and key at the same path.
 func (r *Region) CreateCACertificate(t *testing.T) error {
+	r.CleanUpCACertificate(t)
+
 	// Create CA secret in all regions.
 	cmd := shell.Command{
 		Command:    "cockroach",
--- a/tests/e2e/migrate/controller_migration_test.go
+++ b/tests/e2e/migrate/controller_migration_test.go
@@ -30,6 +30,11 @@ const (
 	v1alpha1CrdbClusterResource  = "crdbclusters.v1alpha1.crdb.cockroachlabs.com"
 	v1beta1CrdbClusterResource   = "crdbclusters.v1beta1.crdb.cockroachlabs.com"
 
+	// Finalization starts after the controller has completed the PodMigration
+	// health checks. Those checks allow up to 10 minutes for a migrated CrdbNode
+	// to become healthy before the controller auto-pauses the migration.
+	phaseFinalizationTimeout = 15 * time.Minute
+
 	// Migration phases as reported in status.migration.phase.
 	phaseCertMigration = "CertMigration"
 	phasePodMigration  = "PodMigration"
@@ -100,7 +105,7 @@ func testHelmAutoForwardMigration(t *testing.T) {
 	// Verify CrdbNodes are being created one by one and cluster stays healthy.
 	waitForCrdbNodeCount(t, kubectlOptions, stsName, crdbCluster.DesiredNodes, 20*time.Minute)
 
-	waitForMigrationPhase(t, kubectlOptions, stsName, phaseFinalization, 5*time.Minute)
+	waitForMigrationPhase(t, kubectlOptions, stsName, phaseFinalization, phaseFinalizationTimeout)
 
 	// Finalization complete: delete the StatefulSet to trigger Phase=Complete.
 	t.Log("Deleting StatefulSet to trigger migration completion")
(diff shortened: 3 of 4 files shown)


shreyaskm623:
Attaching the Screen shot below for the runs of the single and multi region tests which are passing. I have tested them on the `cockroach-shreyaskm` project, I'm sure it should work even in the `helm-testing` project ( where we run the actual weekly test suite ) as these are all project agnostic ( @NishanthNalluri  to confirm ).

Single region :

Screenshot 2026-05-18 at 11 56 21 AM


Multi region : 
[
Screenshot 2026-05-18 at 11 58 08 AM
](url)

tests/e2e/operator/infra/common.go
```diff
@@ -302,8 +302,17 @@ func UpdateKubeconfigGCP(t *testing.T, projectID, region, clusterName, alias str
 		return fmt.Errorf("failed to get GCP credentials for cluster %s: %w", clusterName, err)
 	}
 
-	// Step 2: Rename context
+	// Step 2: Skip TLS verification to avoid x509 errors from GKE's internal CA.
 	longContextName := fmt.Sprintf("gke_%s_%s_%s", projectID, region, clusterName)
+	skipTLSCmd := exec.Command("kubectl", "config", "set-cluster", longContextName,
```
NishanthNalluri: Are we skipping tls verification only while running the tests locally on a GKE cluster or is this the default pattern?

tests/e2e/operator/infra/gcp.go
```diff
@@ -227,6 +235,72 @@ func (r *GcpRegion) SetUpInfra(t *testing.T) {
 	require.NoError(t, err)
 	err = r.deployAndConfigureCoreDNS(t, kubeConfigPath)
 	require.NoError(t, err, "failed to deploy and configure CoreDNS")
+
+	if r.IsMultiRegion {
+		for _, clusterName := range r.Clusters {
+			err = patchKubeDNSForCustomDomains(t, clusterName, kubeConfigPath)
```
NishanthNalluri: Why do we need to patch kubedns with custom cluster domains? Is this mandatory for Multi region tests to work?

tests/e2e/operator/region.go
```diff
@@ -165,6 +171,17 @@ func (r *Region) InstallCharts(t *testing.T, cluster string, index int) {
...
 	helm.Install(t, crdbOptions, helmChartPath, ReleaseName)
 
+	// publishNotReadyAddresses is required on GCP so pods can discover each other
+	// via the join service before they become ready. Skip on k3d — the operator
+	// detects the external drift on a resource it owns and triggers a reconcile
+	// that restarts its own pod, breaking pod readiness on CI runners.
+	if r.Provider == "gcp" {
+		k8s.WaitUntilServiceAvailable(t, kubectlOptions, "cockroachdb-join", 30, 5*time.Second)
+		err := k8s.RunKubectlE(t, kubectlOptions, "patch", "service", "cockroachdb-join",
+			"--type=merge", "-p", `{"spec":{"publishNotReadyAddresses":true}}`)
```
NishanthNalluri: We should not make this change for `join` service, the behaviour should be the same irrespective of provider.

If we install a CRDB cluster manually on a GKE cluster, it works as expected without any modifications right?

tests/e2e/operator/region.go
```diff
@@ -357,6 +374,8 @@ func (r *Region) ValidateCRDBContainerResources(t *testing.T, kubectlOptions *k8
 
 // CreateCACertificate creates CA cert and key at the same path.
 func (r *Region) CreateCACertificate(t *testing.T) error {
+	r.CleanUpCACertificate(t)
```
NishanthNalluri, line 368: Doesnt this create a new CA cert for every test? or Did we refactor to use only one CA for all the tests?

tests/testutil/require.go
```diff
@@ -57,8 +57,8 @@ func RequireClusterToBeReadyEventuallyTimeout(t *testing.T, crdbCluster Cockroac
 		func(ctx context.Context) (bool, error) {
 			ss, err := fetchStatefulSet(crdbCluster.K8sClient, crdbCluster.StatefulSetName, crdbCluster.Namespace)
 			if err != nil {
-				t.Logf("error fetching stateful set")
-				return false, err
+				t.Logf("transient error fetching stateful set, will retry: %v", err)
+				return false, nil
```
NishanthNalluri: why did we make changes to this method? This is not being used by any operator tests right?

NishanthNalluri:
Do we know the exact reason why the existing gcp runs are failing?

tests/e2e/operator/region.go
shreyaskm623: Each test still creates its own CA cert — that behavior hasn't changed. CreateCACertificate is called once at the start of each test function (across both single-region and multi-region tests).                                                                              
 
The CleanUpCACertificate call added at line 377 is purely defensive: it deletes any leftover `ca.crt/ca.key` files before generating a fresh CA. The problem it fixes is that if a previous test crashed mid-run and left stale cert files on disk, the subsequent cockroach cert create-ca command would fail because the files already exist. So it's a "clean slate before creation" guard, not a refactor of the CA lifecycle.

tests/e2e/operator/infra/common.go
shreyaskm623: This is only for tests.

tests/testutil/require.go
shreyaskm623: It's a fix made just for the consistency, Nothing to do with tests.

tests/e2e/operator/infra/gcp.go
shreyaskm623: Yes, For the multi region tests we use `cluster1.local` and `cluster2.local` which needs to be resolved.

Are you suggesting an alternate way here ?

tests/e2e/operator/infra/common.go
NishanthNalluri: No, I meant is this change only for local or are we skipping tls everywhere?

tests/e2e/operator/infra/gcp.go
NishanthNalluri: Even for single region tests we use `cluster1.local` and we don't patch kubedns for those tests right?

tests/e2e/operator/region.go
NishanthNalluri: Thanks, want to confirm this.

tests/testutil/require.go
NishanthNalluri: Yeah, but i want to understand why this change is made now? Did we face any issues with timeout while running tests?

tests/e2e/operator/infra/gcp.go
shreyaskm623: That's exactly why the current tests are failing. Going forward, we'll use `cluster.local` for the single region

You can refer to the screen shot below, after that particular log, the tests will start failing. Even when  I was testing, I faced the similar issue.

Screenshot 2026-05-19 at 3 22 39 PM

tests/e2e/operator/infra/common.go
shreyaskm623: We can skip TLS everywhere for now. In case this requires a holistic fix via operator we can take that up later I think.

tests/testutil/require.go
shreyaskm623: There was an Intermittent issue. It's anyway a very minimal change. I will go ahead and remove it if it's out of scope for this.

tests/e2e/operator/region.go
shreyaskm623: I previously faced the issue with cluster join, that's why had to add this. I tried again without this and works fine now. I have addressed this comment.

tests/e2e/operator/infra/gcp.go
NishanthNalluri: I will check this once because long back the tests are passing with custom domains.

NishanthNalluri: https://github.com/cockroachdb/helm-charts/actions/runs/21127552978 - this is in Jan and we dont have changes to our e2e tests during this time.

shreyaskm623:
@NishanthNalluri For the current fixes in the PR, the build for GCP is passing : https://github.com/cockroachdb/helm-charts/actions/runs/26157873855

NishanthNalluri:
Merging this, will take up rest of the things in upcoming PR's.

PR #636, opened 2026-05-20
e2e/kind: remove unused containerdConfigPatches registry mirror config

shreyaskm623:
Removes an unused containerdConfigPatches block in dev-multi-cluster.sh that referenced an undefined REGISTRY variable, causing the kind cluster creation to fail under set -euo pipefail.
This fixes the kind e2e test failures for both single-region and multi-region.


shreyaskm623:
@NishanthNalluri For the current fixes, both single & multi region tests are passing as expected : https://github.com/cockroachdb/helm-charts/actions/runs/26158434372

NishanthNalluri (approved):
LGTM, Thanks for fixing this!

PR #647, opened 2026-06-18
CC-36481: Add Helm operator troubleshooting and escalation skills

VishalJaishankar:
## Summary
- Add customer-facing CockroachDB Helm skills for install, deployment diagnosis, TLS setup/debugging, and multi-region validation.
- Add a dedicated operator migration debugging skill for Helm StatefulSet and public operator v1alpha1 to v1beta1 migrations.
- Add a separate operator escalation packet collection skill so TSC/TSE artifact gathering does not clutter the normal diagnostic flow.

## Tracking
- Jira: [CC-36481](https://cockroachlabs.atlassian.net/browse/CC-36481)

## Reviewer Context
This split is based on the current TSC/SRE runbooks used with customers:

- https://cockroachlabs.atlassian.net/wiki/spaces/SRE/pages/5464883229/CockroachDB+Operator+Escalation+Debugging+Runbook
- https://cockroachlabs.atlassian.net/wiki/spaces/SRE/pages/5666537508/Operator+Migration+Debugging

The general diagnostic skill stays focused on customer self-service checks and safe next actions. Escalation-only artifacts, pprof/metrics collection, and handoff packaging live in the new escalation packet skill.

## Validation
- Verified the working tree only contained the intended skill files before commit.
- Checked the skill files for trailing whitespace.

### DIFF
--- a/skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
+++ b/skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
@@ -1,0 +1,266 @@
+---
+name: collecting-cockroachdb-operator-escalation-packet
+description: Collects a complete CockroachDB Operator escalation packet for TSC/TSE or operator-team handoff, including Helm state, Kubernetes resources, logs, operation-specific evidence, pprof goroutine dumps, metrics, and a customer action timeline. Use when general diagnosis cannot resolve an operator-managed CockroachDB Helm issue or before restarting a stuck operator.
+compatibility: CockroachDB Helm v2 charts and operator-managed crdb.cockroachlabs.com/v1beta1 resources. Requires Kubernetes read access to the operator namespace and CockroachDB namespace; pprof and metrics collection require port-forward access to the operator Deployment.
+metadata:
+  author: cockroachdb
+  version: "1.0"
+---
+
+# Collecting CockroachDB Operator Escalation Packet
+
+Collects the artifacts needed for TSC/TSE or operator-team escalation. Use this after basic diagnosis identifies an unresolved operator, Kubernetes, migration, upgrade, scale, DNS, PVC, or certificate issue. Keep collection read-only unless the customer explicitly approves a mitigation.
+
+## When to Use This Skill
+
+- A customer needs to escalate an operator-managed CockroachDB Helm issue to TSC
+- The operator is running but not reconciling and logs/status do not explain why
+- A rollout, upgrade, migration, or scale operation is stuck
+- Operator logs do not explain the failure
+- TSC needs consistent artifacts before advising restart, rollback, manual decommission, or status repair
+
+## Safety Considerations
+
+- Collect this packet before restarting, scaling, or deleting the operator during active operations.
+- Do not delete `CrdbNode`, PVC, Secret, service, version checker job, or StatefulSet resources while collecting data.
+- Do not print private key contents. Collect only Secret names, keys, certificate metadata, expiry, issuer, subject, and SANs.
+- Do not manually edit `CrdbCluster.status`.
+- Do not run `cockroach init` on an existing cluster.
+- If the operator is paused or disabled, record that state before changing anything.
+
+## Execution Discipline
+
+- Execute one step at a time and inspect the output before moving on. Do not run the packet collection in parallel; earlier results determine which operation-specific sections are relevant.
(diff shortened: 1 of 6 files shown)


skills/cockroachdb-observability-and-diagnostics/diagnosing-cockroachdb-helm-deployments/SKILL.md
```diff
@@ -0,0 +1,374 @@
...
+
+```bash
+kubectl -n <operator-namespace> rollout restart deploy/cockroach-operator
+```
+
+Force a reconcile:
+
+```bash
+kubectl -n <cockroachdb-namespace> annotate crdbcluster <cockroachdb-release> diagnostics.cockroachlabs.com/kick="$(date +%s)" --overwrite
+```
```
NishanthNalluri, line 370: I'm not sure if we need to this, we already have a no-op timestamp annotation that we use to do rolling-restart. Maybe we could use the same?

skills/cockroachdb-operations-and-lifecycle/configuring-cockroachdb-helm-tls/SKILL.md
```diff
@@ -0,0 +1,261 @@
...
+
+```bash
+kubectl -n <namespace> get certificate,issuer -o wide
+kubectl get clusterissuer -o wide 2>/dev/null || true
+kubectl -n <namespace> describe certificate <certificate-name>
+kubectl -n <namespace> describe issuer <issuer-name>
+```
+
+If the CockroachDB image does not include network or OpenSSL tooling, use an approved debug image according to the customer's policy:
+
```
NishanthNalluri, line 222: Minor nit: This makes sense, we could also add something like for `air-gapped environments` they can mirror the image to their registries or can use something similar which allows has the necessary networking tools.

skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
```diff
@@ -0,0 +1,259 @@
...
+kubectl -n <cockroachdb-namespace> describe pdb > crdb-operator-escalation/pdb-describe.txt
+kubectl -n <cockroachdb-namespace> get events --sort-by=.lastTimestamp > crdb-operator-escalation/events.txt
+```
+
+Summarize key cluster status:
+
+```bash
+kubectl -n <cockroachdb-namespace> get crdbcluster <cockroachdb-release> -o json | jq '{
+  mode: .spec.mode,
+  nodes: .spec.nodes,
```
NishanthNalluri: Should be nodes: [.spec.regions[].nodes]

skills/cockroachdb-observability-and-diagnostics/diagnosing-cockroachdb-helm-deployments/SKILL.md
```diff
@@ -0,0 +1,374 @@
...
+  version: "1.1"
+---
+
+# Diagnosing CockroachDB Helm Deployments
+
+Diagnoses CockroachDB Helm install, upgrade, and readiness failures for operator-managed clusters. Keep the flow customer-facing: collect read-only evidence, classify the failure, and propose the smallest safe remediation. If the issue needs TSC escalation or deep operator forensics, use [collecting-cockroachdb-operator-escalation-packet](../collecting-cockroachdb-operator-escalation-packet/SKILL.md).
+
+## When to Use This Skill
+
+- `helm install`, `helm upgrade`, or `helm spray` fails
```
NishanthNalluri: nit: We can remove helm spray here because it is not supported right now because we are not supporting helm umbrella charts

NishanthNalluri:
This is a great addition that will be extremely helpful for users/our team during cluster deployments and escalations. Thanks @VishalJaishankar 

Couples of things we need to address in the skills as stronger guidance are:

1) We dont want to execute all the commands in parallel (some models tend to do that unless specified), so i think we can mention "Execute each step and review its output before proceeding to the next and sometimes if needed wait for the user confirmation. Earlier steps inform which later steps to run."
2) Never modify anything unless user explicitly specifies (This is because some of the commands exec into shell and also do kubectl patches).
3) If doubtful atleast during escalations or production enviroments (especially to the pprof/metrics collection, mode patching), alway reach to TSE or the team.

I know some of these statments are covered in few places but worth adding stronger guards mentioning for any interpreatation reach out to the team.

NishanthNalluri (approved):
Thanks for adding these skills!

VishalJaishankar:
TFTR!

PR #656, opened 2026-07-17
Fix CRDB skills to discover cluster schema

VishalJaishankar:
## Summary
- add explicit CrdbCluster discovery before object-specific reads
- derive inspected field paths from the live served CRD schema instead of assuming names or stale paths
- update install, migration, TLS, diagnostics, escalation, and multi-region skill flows to capture schema-grounded data
- fix the schema helper for zsh by avoiding the special `path` variable, which mutates `PATH` and breaks `jq` lookup
- stop object-specific drilldown when no live CrdbCluster rows are discovered; remaining CrdbNode owner refs/labels are teardown evidence, not substitute discovery
- add a Codex directive requiring non-empty commit and PR message details for future PRs

## Rationale
The skills should work from observed cluster state first, then drill into discovered CrdbCluster objects using fields exposed by the served CRD schema. This avoids stale assumptions about object names and CRD layouts, and handles teardown states where CrdbNodes remain after the CrdbCluster is gone.

## Validation
- git diff --check
- Markdown code fence balance check
- no remaining unsafe `for path` helper loops
- live OpenShift cluster read using the provided KUBECONFIG confirmed no live CrdbCluster candidates in `cockroach-ns-xxnxyv`
- live CRD schema helper resolved `spec.template.spec.image`, `status.actions`, and `spec.template.spec.certificates` under zsh

### DIFF
--- a/skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
+++ b/skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
@@ -32,13 +32,15 @@ Collects the artifacts needed for TSC/TSE or operator-team escalation. Use this 
 
 - Execute one step at a time and inspect the output before moving on. Do not run the packet collection in parallel; earlier results determine which operation-specific sections are relevant.
 - Keep collection read-only unless the user explicitly approves a mutating action for the target cluster.
+- Never infer a `CrdbCluster` object name from a Helm release, service name, or CockroachDB image/version string. List `CrdbCluster` objects and use the exact `metadata.name`.
+- Before reading version-sensitive `CrdbCluster` fields, save the live CRD YAML and derive field paths from the served CRD schema for the object's `apiVersion`.
 - Do not run interactive `kubectl exec` shells, `kubectl debug`, port-forwards, pprof/metrics collection, or commands that use external images in production unless the user approves them. If impact or policy is unclear, involve TSE or the operator team first.
 - Do not patch, annotate, delete, restart, scale, drain, decommission, or run `helm upgrade` as part of packet collection.
 
 ## Required Inputs
 
 - Operator namespace and Helm release
-- CockroachDB namespace and Helm release
+- CockroachDB namespace, Helm release, and discovered `CrdbCluster.metadata.name`
 - Kubernetes context and cluster/provider
 - Current operation: install, upgrade, scale up/down, cert rotation, migration, routine maintenance, or recovery
 - Current and target CockroachDB image, if an upgrade is involved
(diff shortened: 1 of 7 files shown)


skills/cockroachdb-observability-and-diagnostics/collecting-cockroachdb-operator-escalation-packet/SKILL.md
```diff
@@ -57,64 +59,168 @@ Use clear filenames, for example `operator-logs.txt`, `crdbcluster.yaml`, `event
...
+  actions: value($actionsPath),
+  conditions: value($conditionsPath),
   labels: .metadata.labels,
   annotations: .metadata.annotations
-}' > crdb-operator-escalation/crdbcluster-summary.json
+}' crdb-operator-escalation/crdbcluster.json > crdb-operator-escalation/crdbcluster-summary.json
 
-kubectl -n <cockroachdb-namespace> get crdbnodes \
+kubectl -n "$CRDB_NAMESPACE" get crdbnodes \
   -o custom-columns=NAME:.metadata.name,GENERATION:.metadata.generation,OBSERVED:.status.observedGeneration,PHASE:.status.phase,HASH:.metadata.annotations["crdb.cockroachlabs.com/hash-revision"],NODE_ID:.status.nodeID \
```
NishanthNalluri, line 224: Could you verify this command once? It worked for me after wrapping entire custom-columns in quotes('') and i dont think we have `phase` and `hash-revision` annotation defined in crdbnode status.

skills/cockroachdb-observability-and-diagnostics/diagnosing-cockroachdb-helm-deployments/SKILL.md
```diff
@@ -327,12 +463,12 @@ Common causes:
 ## Scale Down and Decommission
 
 ```bash
-kubectl -n <cockroachdb-namespace> exec <ready-crdb-pod> -c cockroachdb -- \
+kubectl -n "$CRDB_NAMESPACE" exec <ready-crdb-pod> -c cockroachdb -- \
   /cockroach/cockroach node status --decommission
 
-kubectl -n <cockroachdb-namespace> get crdbnodes -o json | jq '[.items[] | select(.status.phase=="Decommissioning")] | {count: length, nodes: [.[].metadata.name]}'
+kubectl -n "$CRDB_NAMESPACE" get crdbnodes -o json | jq '[.items[] | select(.status.phase=="Decommissioning")] | {count: length, nodes: [.[].metadata.name]}'
```
NishanthNalluri, line 469: crdbnode does not have phase field. Should be `select(.status.decommission != null)`.

skills/cockroachdb-observability-and-diagnostics/diagnosing-cockroachdb-helm-deployments/SKILL.md
```diff
@@ -200,17 +282,45 @@ For scoped operators, webhook configurations may be namespace-suffixed, such as
...
+  readyNodes: value($readyNodesPath),
+  statusImage: value($statusImagePath),
+  statusVersion: value($statusVersionPath),
+  actions: value($actionsPath),
+  conditions: value($conditionsPath)
+}' "$CRDB_DIAG_DIR/crdbcluster.json"
+
+kubectl -n "$CRDB_NAMESPACE" get crdbnodes \
   -o custom-columns=NAME:.metadata.name,GENERATION:.metadata.generation,OBSERVED:.status.observedGeneration,PHASE:.status.phase,HASH:.metadata.annotations["crdb.cockroachlabs.com/hash-revision"],NODE_ID:.status.nodeID
 ```
```
NishanthNalluri, line 325: Similar to other comment. `hash-revision` should be `clusterNodeRevision`.

NishanthNalluri:
If some of these issues are originating from runbooks, we should update them as well. In runbooks, make sure commands and helpers are properly indented within code blocks.

Also, since these are skills its no brainer to agents that all of these commands are ran in single shell session but we could add a comment mentioning the same in case if any user tries to run these commands manually.

NishanthNalluri (approved):
Thanks for fixing the buggy commands!

PR #667, opened 2026-08-06
Add OpenShift advanced E2E CI support

VishalJaishankar:
## Summary
- Add OpenShift as an advanced-features E2E provider for both single-region and multi-region when the existing advanced-features workflow is run.
- Add OpenShift installer-backed test infra, including install config generation, kubeconfig merge, OpenShift SCC setup, storage class and Helm value patching, cluster domain handling, and teardown.
- Add multi-region OpenShift setup for Submariner, GCP peer networking, DNS forwarding, debug snapshots, and CI log upload.

## Trigger behavior
This keeps `advanced-features-tests` trigger behavior unchanged from `master`: weekly schedule plus `workflow_dispatch` only. This PR does not add `push` or `pull_request` triggers, so OpenShift runs only when the existing advanced workflow is scheduled or manually dispatched.

## Why this is not just a matrix extension
OpenShift is not a drop-in variant of the existing providers. Extending the matrix starts the jobs, but the shared tests also need provider-specific runtime behavior: cloud provider naming, cluster domains, Helm values, namespace setup, OpenShift `anyuid` SCC binding, Helm uninstall flags, kubeconfig handling, and cleanup.

The provider-runtime interface keeps those differences explicit and out of individual test bodies. That lets the same single-region and multi-region test flows run across k3d, kind, GCP, and OpenShift without spreading provider conditionals through the tests.

The test code changes are limited to the places where the existing shared flow needed provider-specific behavior. The larger OpenShift implementation lives behind the infra/provider boundary rather than extending the matrix with scripts that the Go tests cannot observe or clean up correctly.

## Operational debugging
Where to look:
- GitHub Actions job logs for the failing OpenShift job.
- GitHub artifacts for the captured `test_output.log` and JSON test output.
- GCS debug snapshots under `gs://<GCP_PROJECT_ID>-openshift-ci-debug/<run_id>/<attempt>/<single-region|multi-region>/`.

How to look:
- `gh run view <run_id> --repo cockroachdb/helm-charts --log` for CI logs.
- `gcloud storage ls gs://cockroach-helm-testing-openshift-ci-debug/<run_id>/<attempt>/<mode>/` for uploaded logs and snapshots.
- `gcloud storage cat .../test_output.log` for the Go test stream.
- GCS uploads include `test_output.log`, `test_output.json`, `.openshift_install.log`, `metadata.json`, before/after GCP snapshots, and periodic GCP resource snapshots.

Retrieval notes:
- GitHub logs are the easiest path and are available for all workflow failures.
- GCS debug uploads are best-effort and require project/storage access. If setup fails before the OpenShift test step starts, rely on GitHub logs and artifacts.
- For intentionally preserved clusters, set `OPENSHIFT_SKIP_TEARDOWN=true`; normal scheduled/manual runs leave this false.

Retention:
- CI no longer creates the debug bucket or mutates bucket lifecycle policy. The debug bucket and 30-day retention lifecycle are tracked in cockroachlabs/crl-infrastructure#6230.

## Teardown
Normal scheduled/manual CI runs use `OPENSHIFT_SKIP_TEARDOWN=false`. The Go test cleanup tears down OpenShift clusters and custom multi-region resources. The workflow fallback cleanup now deletes Submariner firewall rules for every known OpenShift infra ID, even when a partial multi-cluster setup only produced one metadata file, then removes peer firewalls and VPC peerings when both sides exist.

## Validation
- `bash -n build/openshift-ci-debug.sh`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/advanced-features-tests.yaml"); puts "yaml ok"'`
- `go test ./tests/e2e/operator/infra ./tests/e2e/operator/multiRegion -run '^$'`
- `git diff --check`

CI evidence:
- OpenShift-only dry run passed: https://github.com/cockroachdb/helm-charts/actions/runs/31000700366
- Temporary full-matrix branch dry run had both OpenShift jobs pass with teardown enabled; the overall run failed in non-OpenShift matrix jobs: https://github.com/cockroachdb/helm-charts/actions/runs/31070731699
- The GCP multi-region failure in that dry run happened inside `TestWALFailoverMultiRegion` after GKE infra came up, and its cleanup completed. The GCP single-region job did not publish retrievable logs/artifacts, so its exact failure point is unavailable from GitHub. These were not OpenShift-provider job failures.
- Temporary branch push runs used while validating this PR were cancelled after restoring trigger behavior.

Pre-PR review follow-up:
- Removed CI-side GCS lifecycle mutation; bucket/lifecycle belongs in Terraform.
- Hardened fallback cleanup for partial multi-cluster setup failures.
- Kept GCS observability explicitly best-effort in the operational notes.

### DIFF
--- a/.github/workflows/advanced-features-tests.yaml
+++ b/.github/workflows/advanced-features-tests.yaml
@@ -6,14 +6,21 @@ on:
   # Allow manual trigger on any branch
   workflow_dispatch:
 
+env:
+  OPENSHIFT_VERSION: "4.17.16"
+  SUBCTL_VERSION: "v0.24.0"
+
 jobs:
   advanced-features-single-region:
     runs-on: [self-hosted, ubuntu_2404]
     strategy:
       fail-fast: false
       matrix:
-        provider: [k3d, kind, gcp]
-    timeout-minutes: 120
+        provider: [k3d, kind, gcp, openshift]
+    timeout-minutes: 300
+    concurrency:
+      group: ${{ matrix.provider == 'openshift' && 'advanced-features-openshift' || format('advanced-features-{0}-{1}-{2}', github.run_id, github.job, matrix.provider) }}
+      cancel-in-progress: false
     permissions:
       contents: read
       id-token: write
@@ -30,7 +37,9 @@ jobs:
         id: auth
         with:
           token_format: access_token
-          project_id: 'cockroach-helm-testing'
+          create_credentials_file: true
+          export_environment_variables: true
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
           workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
       - name: Set up gcloud CLI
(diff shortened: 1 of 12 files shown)


Makefile
```diff
@@ -30,6 +30,13 @@ DOCKER_NETWORK_NAME ?= "k3d-${K3D_CLUSTER}"
 LOCAL_REGISTRY ?= "localhost:5000"
 MULTI_REGION_NODE_SIZE ?= 3
 REGIONS ?= 3
+E2E_OPENSHIFT_SINGLE_REGION_TIMEOUT ?= 180m
+E2E_OPENSHIFT_MULTI_REGION_SETUP_TIMEOUT ?= 300m
+E2E_ADVANCED_SINGLE_REGION_TIMEOUT ?= 90m
+E2E_ADVANCED_MULTI_REGION_TIMEOUT ?= 90m
+E2E_OPENSHIFT_ADVANCED_SINGLE_REGION_TIMEOUT ?= 240m
+E2E_OPENSHIFT_ADVANCED_MULTI_REGION_TIMEOUT ?= 360m
+SUBCTL_VERSION ?= v0.24.0
```
NishanthNalluri, line 35: I think some of these aren't needed if we intend to get rid of additional openshift make commands for openshift.

Makefile
```diff
@@ -146,6 +153,12 @@ test/e2e/multi-region: bin/cockroach bin/kubectl bin/helm  build/self-signer bin
 test/e2e/single-region: bin/cockroach bin/kubectl bin/helm build/self-signer bin/k3d bin/kind
 	@PATH="$(PWD)/bin:${PATH}" go test -timeout 90m -v -test.run TestOperatorInSingleRegion ./tests/e2e/operator/singleRegion/... || (echo "Single region tests failed with exit code $$?" && exit 1)
 
+test/e2e/openshift/single-region: bin/cockroach bin/kubectl bin/helm build/self-signer ## run OpenShift single-region Helm E2E tests
+	@PATH="$(PWD)/bin:${PATH}" PROVIDER=openshift go test -timeout $(E2E_OPENSHIFT_SINGLE_REGION_TIMEOUT) -v -test.run TestOperatorInSingleRegion ./tests/e2e/operator/singleRegion/... || (echo "OpenShift single region tests failed with exit code $$?" && exit 1)
+
+test/e2e/openshift/multi-region-setup: bin/kubectl bin/subctl ## create and validate OpenShift multi-region test infrastructure
+	@PATH="$(PWD)/bin:${PATH}" PROVIDER=openshift go test -timeout $(E2E_OPENSHIFT_MULTI_REGION_SETUP_TIMEOUT) -v -test.run TestOpenShiftMultiRegionInfraSetup ./tests/e2e/operator/multiRegion/... || (echo "OpenShift multi-region setup validation failed with exit code $$?" && exit 1)
+
```
NishanthNalluri: I don’t think we need new make commands specifically for OpenShift. The generic make commands should be sufficient, and the provider can be configured via `PROVIDER`.

The main reason is that adding separate commands for OpenShift sets a precedent for having dedicated commands for every provider.

NishanthNalluri: I get it now. This is added to invoke new infra setup test.

For singlre region, you can invoke with `PROVIDER=openshift make test/e2e/single-region`

Makefile
```diff
@@ -30,6 +30,13 @@ DOCKER_NETWORK_NAME ?= "k3d-${K3D_CLUSTER}"
 LOCAL_REGISTRY ?= "localhost:5000"
 MULTI_REGION_NODE_SIZE ?= 3
 REGIONS ?= 3
+E2E_OPENSHIFT_SINGLE_REGION_TIMEOUT ?= 180m
```
NishanthNalluri: I feel like these timeouts are longer than before. Is it expected that OpenShift cluster provisioning takes more time now? My main concern is that the job might stall for 3 hours with no progress (for example, if there’s an error and the pods don’t come up).

This shouldn’t be an issue, since we have a 30-minute timeout per test. In that case, could we use a common timeout value for every provider?

Makefile
```diff
@@ -216,6 +235,10 @@ bin/kubectl: ## install kubectl
 	@curl -Lo bin/kubectl $(KUBECTL_BIN)
 	@chmod +x bin/kubectl
 
+bin/subctl: ## install subctl
+	@mkdir -p bin
+	@DESTDIR="$(PWD)/bin" VERSION="$(SUBCTL_VERSION)" bash -c 'curl -Ls https://get.submariner.io | bash'
```
NishanthNalluri: nit: just for consistency, can we have this one similar to other binaries.

.github/workflows/advanced-features-tests.yaml
```diff
@@ -30,30 +37,140 @@ jobs:
...
+
+          curl -fsSL "https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/${OPENSHIFT_VERSION}/openshift-client-linux.tar.gz" \
+            -o "${RUNNER_TEMP}/openshift-client-linux.tar.gz"
+          tar -xzf "${RUNNER_TEMP}/openshift-client-linux.tar.gz" -C bin oc kubectl
+
+          chmod +x bin/openshift-install bin/oc bin/kubectl
+
+          if [[ "${OPENSHIFT_ENABLE_SUBMARINER}" == "true" ]]; then
+            DESTDIR="${PWD}/bin" VERSION="${SUBCTL_VERSION}" bash -c 'curl -Ls https://get.submariner.io | bash'
+            bin/subctl version
```
NishanthNalluri: nit: we can just call the make bin/subctl command here.

tests/e2e/operator/multiRegion/openshift_infra_setup_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"github.com/stretchr/testify/require"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const openShiftConnectivityProbeName = "openshift-mr-probe"
+
+// TestOpenShiftMultiRegionInfraSetup validates the OpenShift multi-cluster
+// infrastructure path without installing Helm charts.
+func TestOpenShiftMultiRegionInfraSetup(t *testing.T) {
+	provider := infra.ResolveProvider(t)
```
NishanthNalluri: Isn't this duplicating most of the multi-region infra setup code? I concur on the custom dns probes and any additional checks needed specific to Openshift but in such case, can we add those changes to openshift helpers and invoke them from the existing setup code?

tests/e2e/operator/infra/openshift.go
```diff
@@ -0,0 +1,1580 @@
...
+			deleteOpenShiftFirewallRule(t, projectID, openShiftPeerFirewallName(infraA, infraB))
+			deleteOpenShiftFirewallRule(t, projectID, openShiftPeerFirewallName(infraB, infraA))
+			deleteOpenShiftVPCPeering(t, projectID, openShiftNetworkName(infraA), openShiftPeeringName(infraA, infraB))
+			deleteOpenShiftVPCPeering(t, projectID, openShiftNetworkName(infraB), openShiftPeeringName(infraB, infraA))
+		}
+	}
+}
+
+func (r *OpenShiftRegion) ScaleNodePool(t *testing.T, location string, nodeCount, index int) {
+	if index >= len(r.Clusters) {
```
NishanthNalluri, line 342: Technically, we are not scaling here right? we are provisioning N+1 worker nodes during the initial setup. Having this additional verification is good but this can be a generic helper for all providers? Since we are just checking the ready worker nodes.

Makefile
VishalJaishankar: Addressed in `07ce2a7`: removed the OpenShift-specific make targets and OpenShift-only timeout variables. The OpenShift path now goes through the generic targets with `PROVIDER=openshift`.

Makefile
VishalJaishankar: Addressed in `07ce2a7`: restored the common `90m` Go test timeout for the advanced targets instead of using OpenShift-specific longer test timeouts. The workflow-level timeout remains only as job headroom for provisioning teardown and artifact/debug upload cleanup.

Makefile
VishalJaishankar: Addressed in `07ce2a7`: single-region OpenShift runs through the generic target, e.g. `PROVIDER=openshift make test/e2e/single-region` or the advanced target when running the advanced workflow.

VishalJaishankar: Addressed in `07ce2a7`: removed the dedicated `test/e2e/openshift/...` targets so we do not set a precedent for provider-specific Make targets. The generic targets are used with `PROVIDER=openshift`.

tests/e2e/operator/infra/openshift.go
VishalJaishankar: Addressed in `07ce2a7`: moved the ready-node polling into a generic `infra.WaitForReadyNodes` helper and changed OpenShift `ScaleNodePool` to call it with the worker-node selector.

.github/workflows/advanced-features-tests.yaml
VishalJaishankar: Addressed in `07ce2a7`: the workflow now calls `make bin/subctl` for OpenShift multi-region instead of installing Submariner inline.

tests/e2e/operator/multiRegion/openshift_infra_setup_test.go
VishalJaishankar: Addressed in `07ce2a7`: removed the standalone `openshift_infra_setup_test.go` instead of duplicating the multi-region setup flow. The OpenShift-specific DNS forwarding validation now lives in the OpenShift infra helper and runs from the shared setup path.

Makefile
VishalJaishankar: Addressed in `07ce2a7`: changed `bin/subctl` to match the other binary targets by downloading the versioned release archive via `SUBCTL_BIN`, extracting into `bin/`, and marking it executable.

.github/workflows/advanced-features-tests.yaml
```diff
@@ -6,14 +6,21 @@ on:
   # Allow manual trigger on any branch
   workflow_dispatch:
 
+env:
+  OPENSHIFT_VERSION: "4.17.16"
+  SUBCTL_VERSION: "v0.24.0"
+
```
NishanthNalluri, line 12: nit: I understand the intention here - to override custom version. Just flagging at this point its redundant.

tests/e2e/operator/infra/common.go
```diff
@@ -384,3 +390,58 @@ func UpdateCoreDNSWithNamespaces(t *testing.T, r *operator.Region) {
...
+		}
+
+		readyNodes := countReadyNodes(output)
+		t.Logf("[%s] ready nodes for cluster %q: %d/%d",
+			provider, clusterName, readyNodes, nodeCount)
+		if readyNodes >= nodeCount {
+			return
+		}
+
+		time.Sleep(defaultRetryInterval)
```
NishanthNalluri, line 427: nit: we could replace this with backoff/retry (i think the similar pattern is used in other places).

NishanthNalluri (approved):
LGTM

PR #672, opened 2026-08-21
ci: add Azure E2E infrastructure

VishalJaishankar:
## Summary

Adds Azure-backed E2E infrastructure and CI wiring for the Helm chart operator test suites.

This PR is for [CC-35912](https://cockroachlabs.atlassian.net/browse/CC-35912).

## Changes

- Adds an AKS-backed Azure provider for the operator E2E harness.
- Provisions tagged Azure resource groups, VNets/subnets, AKS clusters, VNet peerings, CoreDNS service/configuration, and Azure-specific provider runtime behavior.
- Wires Azure into the integration and advanced-features workflows for single-region, multi-region, and advanced E2E runs.
- Adds Azure workflow authentication with `azure/login`, Azure-specific timeouts, and branch-scoped concurrency groups so Azure jobs do not overlap for the same branch/test class.
- Records created Azure resource groups through `AZURE_RESOURCE_GROUPS_FILE` so cleanup can run independently from Go test teardown.
- Tags Azure resource groups with `ManagedBy`, `Ticket`, `TestRun`, GitHub run metadata, workflow, job, and ref name for ownership and stale-resource detection.

## Azure CI Features

- Supports `PROVIDER=azure` for operator single-region and multi-region E2E suites.
- Runs the Azure matrix entries in both scheduled and manually triggered CI workflows.
- Uses generated per-run resource names, isolated kubeconfigs, and Azure-aware Kubernetes retry behavior.
- Configures CoreDNS for Azure cluster domains and multi-region DNS forwarding.
- Uses async final Azure resource-group cleanup to keep successful CI runs fast.
- Performs a pre-run stale-resource check before provisioning new Azure infrastructure. If stale groups from previous runs exist, the job submits async cleanup for them and fails before creating new resources.

## Debugging And O11y

`build/azure-ci-debug.sh` captures Azure CI snapshots before stale cleanup, before/after tests, periodically during long-running tests, and before/after cleanup.

Each snapshot includes:

- Azure CLI account/subscription context.
- Recorded resource groups for the current run.
- Known `helm-charts` Azure resource groups by prefix/tag.
- Resource group details and resource inventory.
- AKS clusters, provisioning states, power states, Kubernetes versions, and AKS node resource groups.
- VNets, VNet peerings, public IPs, load balancers, disks, and recent failed Azure activity-log events.

To access the debug output:

1. Open the GitHub Actions run.
2. Open the Azure job, such as `advanced-features-single-region (azure)` or `advanced-features-multi-region (azure)`.
3. Download the uploaded job artifact, for example `advanced-single-region-results-azure` or `advanced-multi-region-results-azure`.
4. Inspect `_temp/azure-debug/azure-*.log` and `_temp/azure-resource-groups.txt` inside the artifact.

A validation run used while developing this CI is [GitHub Actions run 32354470589](https://github.com/cockroachdb/helm-charts/actions/runs/32354470589).

## Cleanup Model

Cleanup is intentionally split into two paths:

- End-of-run cleanup submits async `az group delete --no-wait` for the resource groups recorded during the test run.
- Start-of-run preflight checks for stale tagged resource groups from prior runs. If any are found, the workflow submits async cleanup and fails immediately before provisioning anything new.

This keeps normal successful runs fast while making leftover Azure resources visible and actionable instead of silently carrying them into a new run.

## Validation

- `bash -n build/azure-ci-debug.sh`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts f }' .github/workflows/advanced-features-tests.yaml .github/workflows/integration-tests.yaml`
- `git diff --check HEAD`
- `go test ./tests/e2e/operator/infra`
- `go test ./tests/e2e/operator/... -run '^$'`
- Live Azure query confirmed no matching `hc-e2e`, `helm-charts-e2e`, `MC_hc-e2e`, or `ManagedBy=helm-charts-e2e` resource groups/resources remained after [GitHub Actions run 32354470589](https://github.com/cockroachdb/helm-charts/actions/runs/32354470589).

`actionlint` was not available in the local environment, so workflow validation was limited to YAML parsing and the checks above.

### DIFF
--- a/.github/workflows/advanced-features-tests.yaml
+++ b/.github/workflows/advanced-features-tests.yaml
@@ -16,10 +16,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        provider: [k3d, kind, gcp, openshift]
+        provider: [k3d, kind, gcp, openshift, azure]
     timeout-minutes: 300
     concurrency:
-      group: ${{ matrix.provider == 'openshift' && 'advanced-features-openshift' || format('advanced-features-{0}-{1}-{2}', github.run_id, github.job, matrix.provider) }}
+      group: ${{ matrix.provider == 'azure' && format('helm-charts-azure-e2e-{0}-advanced-single-region', github.ref_name) || matrix.provider == 'openshift' && 'advanced-features-openshift' || format('advanced-features-{0}-{1}-{2}', github.run_id, github.job, matrix.provider) }}
       cancel-in-progress: false
     permissions:
       contents: read
@@ -33,6 +33,7 @@ jobs:
         with:
           go-version: 1.26
       - name: Authenticate to Google Cloud
+        if: ${{ matrix.provider == 'gcp' || matrix.provider == 'openshift' }}
         uses: 'google-github-actions/auth@v2'
         id: auth
         with:
@@ -43,10 +44,43 @@ jobs:
           workload_identity_provider: ${{ secrets.GCP_WORKLOAD_IDENTITY_PROVIDER }}
           service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
       - name: Set up gcloud CLI
+        if: ${{ matrix.provider == 'gcp' || matrix.provider == 'openshift' }}
         uses: google-github-actions/setup-gcloud@v2
       - name: Install gke-gcloud-auth-plugin
+        if: ${{ matrix.provider == 'gcp' || matrix.provider == 'openshift' }}
         run: |
           gcloud components install gke-gcloud-auth-plugin
+      - name: Authenticate to Azure
+        if: ${{ matrix.provider == 'azure' }}
+        uses: azure/login@v2
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+      - name: Verify Azure CLI
+        if: ${{ matrix.provider == 'azure' }}
+        shell: bash
+        env:
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+        run: |
+          set -euo pipefail
+          az version
+          az account show --subscription "${AZURE_SUBSCRIPTION_ID}" --output table
+      - name: Clean up stale Azure resource groups
+        if: ${{ matrix.provider == 'azure' }}
+        shell: bash
+        env:
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          AZURE_RESOURCE_PREFIX: hc-e2e
+          AZURE_TICKET: CC-35912
+          AZURE_DEBUG_DIR: ${{ runner.temp }}/azure-debug
+        run: |
+          set -euo pipefail
+          ./build/azure-ci-debug.sh snapshot before-stale-cleanup
+          stale_status=0
+          ./build/azure-ci-debug.sh cleanup-old-resource-groups || stale_status="$?"
+          ./build/azure-ci-debug.sh snapshot after-stale-cleanup
+          exit "${stale_status}"
       - name: Install OpenShift tools
         if: ${{ matrix.provider == 'openshift' }}
         shell: bash
@@ -82,7 +116,7 @@ jobs:
             echo "EOF"
           } >> "${GITHUB_ENV}"
       - name: Run tests (advanced features - single region)
-        if: ${{ matrix.provider != 'openshift' }}
+        if: ${{ matrix.provider != 'openshift' && matrix.provider != 'azure' }}
         env:
           PROVIDER: ${{ matrix.provider }}
           TEST_ADVANCED_FEATURES: true
@@ -90,6 +124,35 @@ jobs:
         run: |
           set -euo pipefail
           make test/nightly-e2e/advanced/single-region | tee test_output.log
+      - name: Run tests (advanced features - single region - Azure)
+        if: ${{ matrix.provider == 'azure' }}
+        shell: bash
+        env:
+          PROVIDER: ${{ matrix.provider }}
+          TEST_ADVANCED_FEATURES: true
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          AZURE_RESOURCE_PREFIX: hc-e2e
+          AZURE_TICKET: CC-35912
+          AZURE_RESOURCE_GROUPS_FILE: ${{ runner.temp }}/azure-resource-groups.txt
+          AZURE_DEBUG_DIR: ${{ runner.temp }}/azure-debug
+          E2E_ADVANCED_SINGLE_REGION_TEST_TIMEOUT: 120m
+        run: |
+          set -euo pipefail
+
+          ./build/azure-ci-debug.sh snapshot before-single-region
+          ./build/azure-ci-debug.sh watch 120 &
+          debug_pid="$!"
+          finish_debug() {
+            status="$?"
+            kill "${debug_pid}" 2>/dev/null || true
+            wait "${debug_pid}" 2>/dev/null || true
+            ./build/azure-ci-debug.sh snapshot after-single-region || true
+            exit "${status}"
+          }
+          trap finish_debug EXIT
+
+          unset AZURE_CLIENT_ID AZURE_CLIENT_SECRET AZURE_TENANT_ID AZURE_REUSE_CONTEXTS AZURE_RESOURCE_GROUP
+          stdbuf -oL -eL make test/nightly-e2e/advanced/single-region 2>&1 | tee test_output.log
       - name: Run tests (advanced features - single region - OpenShift)
         if: ${{ matrix.provider == 'openshift' }}
         env:
@@ -122,6 +185,22 @@ jobs:
 
           unset OPENSHIFT_REUSE_CONTEXTS OPENSHIFT_REUSE_KUBECONFIG OPENSHIFT_REUSE_KUBECONFIGS
           stdbuf -oL -eL make test/nightly-e2e/advanced/single-region 2>&1 | tee test_output.log
+      - name: Clean up Azure resource groups
+        if: ${{ always() && matrix.provider == 'azure' }}
+        shell: bash
+        env:
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          AZURE_TICKET: CC-35912
+          AZURE_RESOURCE_PREFIX: hc-e2e
+          AZURE_RESOURCE_GROUPS_FILE: ${{ runner.temp }}/azure-resource-groups.txt
+          AZURE_DEBUG_DIR: ${{ runner.temp }}/azure-debug
+        run: |
+          set -euo pipefail
+          ./build/azure-ci-debug.sh snapshot cleanup-before-single-region || true
+          cleanup_status=0
+          ./build/azure-ci-debug.sh cleanup-resource-groups || cleanup_status="$?"
+          ./build/azure-ci-debug.sh snapshot cleanup-after-single-region || true
+          exit "${cleanup_status}"
       - name: Archive test results
         if: ${{ always() }}
         uses: actions/upload-artifact@v4
@@ -132,6 +211,8 @@ jobs:
           path: |
             test_output.log
             test_output.json
+            ${{ runner.temp }}/azure-debug/*.log
+            ${{ runner.temp }}/azure-resource-groups.txt
             ${{ runner.temp }}/openshift-debug/*.log
             ${{ runner.temp }}/helm-charts-ocp-*/.openshift_install.log
             ${{ runner.temp }}/helm-charts-ocp-*/metadata.json
@@ -179,10 +260,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        provider: [k3d, kind, gcp, openshift]
+        provider: [k3d, kind, gcp, openshift, azure]
     timeout-minutes: 420
     concurrency:
-      group: ${{ matrix.provider == 'openshift' && 'advanced-features-openshift' || format('advanced-features-{0}-{1}-{2}', github.run_id, github.job, matrix.provider) }}
+      group: ${{ matrix.provider == 'azure' && format('helm-charts-azure-e2e-{0}-advanced-multi-region', github.ref_name) || matrix.provider == 'openshift' && 'advanced-features-openshift' || format('advanced-features-{0}-{1}-{2}', github.run_id, github.job, matrix.provider) }}
       cancel-in-progress: false
     permissions:
       contents: read
@@ -196,6 +277,7 @@ jobs:
         with:
           go-version: 1.26
       - name: Authenticate to Google Cloud
+        if: ${{ matrix.provider == 'gcp' || matrix.provider == 'openshift' }}
         uses: 'google-github-actions/auth@v2'
         id: auth
         with:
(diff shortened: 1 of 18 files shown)


build/azure-ci-debug.sh
```diff
@@ -0,0 +1,355 @@
...
+      return 1
+      ;;
+    *)
+      echo "Unexpected Azure resource group existence response for ${rg}: ${exists}" >&2
+      return 2
+      ;;
+  esac
+}
+
+wait_resource_group_deleted() {
```
NishanthNalluri: Is this being called somewhere? i could not see in git workflow or in this script.

build/azure-ci-debug.sh
```diff
@@ -0,0 +1,355 @@
...
+  local rg
+  for rg in "${groups[@]}"; do
+    [[ -n "${rg}" ]] || continue
+    delete_resource_group_async "${rg}" || cleanup_status="$?"
+  done
+  return "${cleanup_status}"
+}
+
+cleanup_old_resource_groups() {
+  if [[ "${AZURE_CLEANUP_OLD_RESOURCE_GROUPS:-true}" != "true" ]]; then
```
NishanthNalluri: Are these environment variables intended to be defined in the workflow?

.github/workflows/advanced-features-tests.yaml
```diff
@@ -82,14 +116,43 @@ jobs:
...
           make test/nightly-e2e/advanced/single-region | tee test_output.log
+      - name: Run tests (advanced features - single region - Azure)
+        if: ${{ matrix.provider == 'azure' }}
+        shell: bash
+        env:
+          PROVIDER: ${{ matrix.provider }}
+          TEST_ADVANCED_FEATURES: true
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}
+          AZURE_RESOURCE_PREFIX: hc-e2e
+          AZURE_TICKET: CC-35912
```
NishanthNalluri: nit: we should not mention jira ticket here.

build/azure-ci-debug.sh
```diff
@@ -0,0 +1,355 @@
...
+  local interval="${1:-120}"
+  while true; do
+    snapshot "periodic"
+    sleep "${interval}"
+  done
+}
+
+cleanup_resource_groups() {
+  if [[ "${AZURE_SKIP_TEARDOWN:-false}" == "true" ]]; then
+    echo "AZURE_SKIP_TEARDOWN=true; preserving recorded Azure resource groups"
```
NishanthNalluri: Same with this env var; Is our plan to skip teardown only when the tests fail? If so, we can set it to false in the workflow.

.github/workflows/integration-tests.yaml
```diff
@@ -26,24 +29,103 @@ jobs:
...
           service_account: ${{ secrets.GCP_SERVICE_ACCOUNT }}
       - name: Set up gcloud CLI
+        if: ${{ matrix.provider == 'gcp' }}
         uses: google-github-actions/setup-gcloud@v2
       - name: Install gke-gcloud-auth-plugin
+        if: ${{ matrix.provider == 'gcp' }}
         run: |
           gcloud components install gke-gcloud-auth-plugin
+      - name: Authenticate to Azure
+        if: ${{ matrix.provider == 'azure' }}
```
NishanthNalluri, line 47: With addition of new cloud providers, these workflows have become beefy. I think it would be helpful to move the duplicated logic into actions and call them wherever needed.

tests/e2e/operator/provider_runtime.go
```diff
@@ -62,10 +68,33 @@ func (r *Region) clusterDomain(index int) string {
...
+		if apierrors.IsAlreadyExists(err) {
+			return "", nil
+		}
+		return "", err
+	})
+	require.NoError(t, err)
 	r.runtime().ConfigureNamespace(t, kubectlOptions, namespace)
 }
 
+func (r *Region) shouldRetryKubernetesAction(opts *k8s.KubectlOptions) bool {
```
NishanthNalluri, line 88: i dont think i understand this, why do we need a retries only for azure provider? If we encounter any transient issues during namespace creation, we could use existing retry pattern (even for the other helm/k8s ops).

tests/e2e/operator/region.go
```diff
@@ -533,6 +587,118 @@ func HelmChartPaths() (helmChartPath string, operatorChartPath string) {
...
+func HelmInstallWithRetry(t *testing.T, options *helm.Options, chart, releaseName string) {
+	t.Helper()
+	if !shouldRetryHelmAction(options) {
+		helm.Install(t, options, chart, releaseName)
+		return
+	}
+
+	err := runHelmActionWithRetry(t, fmt.Sprintf("helm install %s", releaseName), func() error {
+		err := helm.InstallE(t, options, chart, releaseName)
+		if isHelmInstallAlreadyExists(err) {
```
NishanthNalluri, line 599: I feel that all these retry patterns are redundant. I believe they may have been added during local testing, where consecutive runs could have resulted in such issues.

Ideally, in these cases, we should tear down/uninstall at the beginning of test run. CI is unlikely to encounter these issues because the tests run in a new VPC almost every time.

tests/e2e/operator/region.go
```diff
@@ -533,6 +587,118 @@ func HelmChartPaths() (helmChartPath string, operatorChartPath string) {
...
+		err := helm.DeleteE(t, options, releaseName, purge)
+		if isHelmReleaseNotFound(err) {
+			return nil
+		}
+		return err
+	})
+}
+
+func shouldRetryHelmAction(options *helm.Options) bool {
+	if strings.EqualFold(strings.TrimSpace(os.Getenv("PROVIDER")), "azure") {
```
NishanthNalluri, line 640: Similar to the one mentioned [here](https://github.com/cockroachdb/helm-charts/pull/672/changes#r3861592176).

I dont think these retry patterns needs to be reinvented as we already have few helpers from upstream pkgs or in this codebase.

tests/e2e/operator/region.go
```diff
@@ -533,6 +587,118 @@ func HelmChartPaths() (helmChartPath string, operatorChartPath string) {
...
+		}
+		if !isTransientHelmError(err) {
+			return err
+		}
+		if attempt == attempts {
+			break
+		}
+		t.Logf("%s failed with a transient Kubernetes API error on attempt %d/%d: %v", description, attempt, attempts, err)
+		time.Sleep(20 * time.Second)
+	}
```
NishanthNalluri, line 667: This helper also seems similar to backoff/retry logic. I’m sure we can plug custom errors without having to implement the entire retry/backoff mechanism ourselves.

tests/testutil/require.go
```diff
@@ -439,6 +484,28 @@ func RequireCertificatesToBeValid(t *testing.T, crdbCluster CockroachCluster) {
 	t.Log("Certificates validated successfully")
 }
 
+func getSecretWithRetry(t *testing.T, kubectlOptions *k8s.KubectlOptions, secretName string) *corev1.Secret {
+	if !shouldRetryKubernetesContext(kubectlContext(kubectlOptions)) {
```
NishanthNalluri, line 488: having retry on every resource is a good pattern but we could make most of the generic.

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+		DNSServiceIP: "172.28.17.10",
+	},
+	{
+		Region:       "westus2",
+		VNetCIDR:     "10.20.0.0/16",
+		SubnetCIDR:   "10.20.0.0/24",
+		ServiceCIDR:  "172.28.49.0/24",
+		DNSServiceIP: "172.28.49.10",
+	},
+}
```
NishanthNalluri: This config is already defined in the common.go, so this must be redundant?

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+		"UNEXPECTED_EOF_WHILE_READING",
+		"EOF occurred in violation of protocol",
+		"Certificate verification failed",
+		"Connection aborted",
+		"connection reset",
+		"context deadline exceeded",
+		"timed out",
+		"TLS/SSL connection has been closed",
+		"unexpected EOF",
+	} {
```
NishanthNalluri, line 700: Are all these errors expected during the actual CI run?

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+		Data: buildAzureCoreDNSCustomData(thisDomain, allClusters),
+	}
+	if err := kubectlApplyAzureManifest(t, kubectlOpts, coredns.ToYAML(t, cm)); err != nil {
+		return fmt.Errorf("kubectl apply coredns-custom: %w", err)
+	}
+	return nil
+}
+
+func buildAzureCoreDNSCustomData(
+	thisDomain string, allClusters map[string]coredns.CoreDNSClusterOption,
```
NishanthNalluri: Isn’t this similar to the custom CoreDNS helper we have?

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+			data[fmt.Sprintf("%s.pod.%s.server", cluster.Namespace, clusterDomain)] =
+				serverBlock(fmt.Sprintf("%s.pod.%s", cluster.Namespace, clusterDomain))
+		}
+	}
+
+	return data
+}
+
+func applyAzureCoreDNSService(t *testing.T, kubectlOpts *k8s.KubectlOptions) error {
+	annotations := GetLoadBalancerAnnotations(ProviderAzure)
```
NishanthNalluri: The same applies to this method: we already have these helper methods, so we can reuse and extend them as needed.

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+		"HTTP_PROXY",
+		"HTTPS_PROXY",
+		"NO_PROXY",
+		"http_proxy",
+		"https_proxy",
+		"no_proxy",
+		"REQUESTS_CA_BUNDLE",
+		"AZURE_CLI_CA_CERTS",
+		"SSL_CERT_FILE",
+	} {
```
NishanthNalluri, line 82: Are these mostly needed while running tests locally?

tests/e2e/operator/infra/azure.go
```diff
@@ -0,0 +1,1242 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	if err := cmd.Run(); err != nil {
+		return fmt.Errorf("az network vnet create %s: %w", cfg.VNetName, err)
+	}
+	return nil
+}
+
+func (r *AzureRegion) createAKSClusters(t *testing.T) error {
+	for i, clusterName := range r.Clusters {
```
NishanthNalluri, line 526: This creates the AKS clusters serially for MR clusters right? We could parallelise it similar to other providers.

.github/workflows/integration-tests.yaml
VishalJaishankar: Agreed the workflows have grown with each new provider. I'd like to keep this PR scoped to adding the Azure matrix entry so it doesn't collide with in-flight branches touching the same workflows. Happy to open a follow-up PR that extracts the shared k3d/kind/gcp/openshift/azure setup into composite actions.

tests/e2e/operator/provider_runtime.go
VishalJaishankar: The inner call is `terratest/modules/retry.DoWithRetryE`, so no bespoke retry loop. The wrapper exists to (a) gate retries behind `shouldRetryKubernetesAction` so k3d/kind/gcp behavior is unchanged from before this PR, and (b) short-circuit on `apierrors.IsAlreadyExists` so a re-attempted namespace create is not a hard failure. AKS's kube-apiserver was empirically the only one where `CreateNamespace` hit transient EOF/TLS errors during the validation run linked in the PR description.

tests/e2e/operator/region.go
VishalJaishankar: These paths are guarded by `shouldRetryHelmAction`, which only activates for Azure — in every other provider the call site is `helm.Install`/`helm.Delete` exactly as before this PR, no wrapper. So there's no CI regression on the assumption of a fresh VPC.

The Azure-only wrapper exists because during multi-region AKS runs the kube-apiserver drops connections in ways that don't match the errors terratest's retries look for by default (e.g. `IncompleteCertChain`, `SSL_INCOMPLETE_CHAIN`, `TLS handshake timeout` at helm plan time). The substance is the classifier `isTransientHelmError`; the loop itself is a thin `for` — happy to swap it for `retry.DoWithRetryE` as a follow-up, but the classifier is what pays for itself.

tests/e2e/operator/region.go
VishalJaishankar: Same story as https://github.com/cockroachdb/helm-charts/pull/672#discussion_r3861631301 — the loop uses a custom for because it needs to probe both `isHelmInstallAlreadyExists` (delete-then-retry) and `isTransientHelmError` (retry as-is) on each error, which `DoWithRetryE` cannot express without unwrapping. Only activates for Azure so non-Azure providers are unchanged.

tests/e2e/operator/region.go
VishalJaishankar: This one isn't a generic backoff — it's specifically for the AKS async provisioning state machine (`Succeeded` / `Failed` / `Canceled` / empty / other). It's called only after `az aks create` returns a transient CLI transport error, to check whether the cluster actually reached `Succeeded` despite the CLI dropping the connection. Generic retry would just re-run `az aks create` (~15 min) instead of polling `az aks show` (~seconds).

tests/testutil/require.go
VishalJaishankar: Agree it'd be nice to generalize. Kept Azure-scoped in this PR so that (a) k3d/kind/gcp/openshift behavior is byte-for-byte unchanged for reviewers to reason about, and (b) each provider's transient-error classifier gets validated independently before we start hiding failures behind retries across the board. I'd rather do the generic refactor as a follow-up PR where the risk/benefit is the whole point of the change.

tests/e2e/operator/infra/azure.go
VishalJaishankar: Every marker was observed at least once during the validation run linked in the PR description (GH Actions run 32354470589). AKS control-plane restarts produce `TLS handshake timeout`, `unexpected EOF`, `Connection aborted`, `EOF occurred in violation of protocol` on separate steps of the same run. Removing any of them just re-introduces flakes.

tests/e2e/operator/infra/azure.go
VishalJaishankar: Yes, primarily for local dev when running behind a corporate proxy — we preserve `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY`/cert-bundle vars for `az` even after `disableKubernetesProxyEnv` unsets them for kubectl (kubectl in-cluster traffic must not go via proxy, but `az` public-endpoint traffic must). In CI these env vars are unset, so `captureAzureCLIEnv` returns an empty map and the override is a no-op. Zero-cost in CI, useful for local dev.

build/azure-ci-debug.sh
VishalJaishankar: Removed in 8669edd — this function (along with its only caller `resource_group_exists` and the orphaned `cleanup_wait_seconds`) was dead code.

build/azure-ci-debug.sh
VishalJaishankar: Removed the `AZURE_CLEANUP_OLD_RESOURCE_GROUPS` gate in 8669edd — the flag defaulted to `true` and no workflow ever set it. Stale cleanup now always runs.

.github/workflows/advanced-features-tests.yaml
VishalJaishankar: Dropped `AZURE_TICKET` entirely in 8669edd — env var, `Ticket=` resource-group tag, and the ticket-based filter path. Stale cleanup still segregates runs via `ManagedBy`, `GitHubWorkflow`, `GitHubJob`, `GitHubRefName`, and `GitHubRunID != current`.

build/azure-ci-debug.sh
VishalJaishankar: Removed the `AZURE_SKIP_TEARDOWN` gate in 8669edd — nothing set it in the workflow, teardown always runs now. If someone needs to preserve resources for local debugging they can do it manually.

tests/e2e/operator/infra/azure.go
VishalJaishankar: Fixed in 8669edd — `azureClusterConfigTemplates` deleted. `buildClusterConfigs` now reads the region's VNet/Subnet/Service/DNS values from `NetworkConfigs[ProviderAzure]` in `common.go`.

tests/e2e/operator/infra/azure.go
VishalJaishankar: Refactored in 8669edd. The shared `coredns.CoreDNSConfigMap` was already computing the `.server` / `.override` map entries (which AKS's `coredns-custom` needs) but discarding them. Exposed them via a new `coredns.CoreDNSCustomConfigMapData`; `applyAzureCoreDNSCustom` now calls that and the duplicate `buildAzureCoreDNSCustomData` is gone.

tests/e2e/operator/infra/azure.go
VishalJaishankar: Fixed in 8669edd — `applyAzureCoreDNSService` now uses `coredns.CoreDNSService(nil, annotations)` and only overrides the selector (AKS may use `k8s-app: coredns` instead of `kube-dns`, detected at runtime via `detectAzureCoreDNSPodLabel`).

tests/e2e/operator/infra/azure.go
VishalJaishankar: Parallelized in 8669edd — the two `az aks create` calls now run concurrently via `sync.WaitGroup`, so MR runs finish in ~15 min instead of ~30. Kubeconfig writes remain serial after both clusters return, to avoid concurrent writers to the shared `KUBECONFIG` file.

NishanthNalluri (approved):
Thanks for addressing the review comments and fixing the nits. Once we are done with all provider implementations, we can refactor the code and remove the redundant paths in a follow-up PR.

PR #676, opened 2026-09-13
release: prepare operator v1.1.0 release

NishanthNalluri:
This PR advances the operator chart and its related images to v1.1.0. The complete list of changes and fixes is included in the changelog.

Key changes:
- Improves observability and recovery when cluster initialization hangs.
- Removes `CrdbCluster` ownership from cluster-scoped `PriorityClasses`.
- Fixes the under-replicated ranges check to use a ready CockroachDB pod.
- Prevents WAL volume resizes from triggering rolling restarts.
- Prevents the operator from running without a region detected from node labels or supplied through the `CLOUD_REGION` environment variable.
- Fixes topology-aware scale-down when `topologySpreadConstraints` are configured through `podTemplate`.
- Keeps the migration conversion webhook CA bundle synchronized after operator certificate rotation.

Until the operator is upgraded to v1.1.0, users who moved `topologySpreadConstraints` into `podTemplate` must continue setting the equivalent deprecated first-class field. It can be removed after all operator replicas are running v1.1.0.

Additionally, updates the K3s test image to v1.32.13-k3s1.


NishanthNalluri:
Superseded by #680 after renaming the release branch to `nishanth/operator-v1.1.0-release`.

PR #678, opened 2026-09-16
rename chart folders: cockroachdb-legacy + cockroachdb-operator

VishalJaishankar:
Refs [CC-36950](https://cockroachlabs.atlassian.net/browse/CC-36950).

## Rename

| Before | After |
|---|---|
| `cockroachdb/` | `cockroachdb-legacy/` |
| `cockroachdb-parent/` | `cockroachdb-operator/` |

Published chart artifact names (`cockroachdb`, `cockroachdb-chart`, `cockroachdb-operator-chart`) are unchanged; helm-repo consumers are unaffected.

## Umbrella chart `name:` change (customer-visible, but no pod restarts)

The umbrella chart's `Chart.yaml` `name:` field flips from `cockroachdb-parent` → `cockroachdb-operator` to match the folder.

**Does not restart pods**: the umbrella chart has no `templates/` of its own. Every rendered resource's `helm.sh/chart` label is set from the subchart-scoped `.Chart.Name` via the `cockroachdb.chart` helper — those resolve to `cockroachdb-<version>` and `cockroachdb-operator-chart-<version>` and are unchanged by the umbrella rename. Verified by rendering the umbrella and inspecting every `helm.sh/chart` label.

**What does change**:
- `helm list` CHART column: `cockroachdb-parent-<version>` → `cockroachdb-operator-<version>`
- Release-secret `chart.metadata.name` is rewritten on next upgrade
- Any tooling that filters helm releases by chart name (`helm list --filter '^cockroachdb-parent'`, GitOps drift detection keyed on chart name) needs to be updated

## Customer impact

- **Helm-repo installs** (`helm install cockroachdb/cockroachdb-chart`): no change.
- **Git-clone installs**: `./cockroachdb-parent` → `./cockroachdb-operator`, `./cockroachdb` → `./cockroachdb-legacy`. Documented in `CHANGELOG.md`.
- **Legacy chart README**: deprecation banner + link to the [Kubernetes deprecation notice](https://www.cockroachlabs.com/docs/v26.2/kubernetes-deprecation-notice).

## Historical git tags

The historical `[cockroachdb-parent-<version>]` headings preserved in `CHANGELOG.md` are documentation labels — no git tag ever used that prefix. The corresponding release commits are tagged `v<version>-preview+<n>` in git. `MIGRATION_v1alpha1_to_v1beta1.md` is updated to use the real `v*` tag names, with a preamble note explaining that helm paths inside that guide intentionally reference the pre-rename `cockroachdb-parent/` folder (that's what exists at those historical revisions).

## Scope

158 files, single commit. Touches `build/**`, `.github/workflows/{ci,automerge}.yaml`, `Makefile`, `scripts/**`, `docs/**`, `tests/**`, `skills/**`, `examples/**`, and the two renamed chart trees.

## Validation

`make validate/rename[/quick]` — 8 steps: static grep guards (both trailing-file paths, terminal-token Go string literals, AND bare `./cockroachdb` shell/markdown uses), old-folder-absent, new-folder-present, generate idempotency (hash-stable), `helm lint`, `helm dependency update`, `go test ./build/... ./tests/template/...`, k3d smoke deploy of both charts.

Steps 1–7 pass locally. E2E on fresh k3d (cache purged):
- Legacy chart from `./cockroachdb-legacy`: 3 CRDB pods Ready, `SELECT 1` = 1, `cockroach node status` shows 3 live nodes on v26.3.1.
- Operator umbrella from `./cockroachdb-operator`: helm install ok, CRDs applied, operator reconciles CrdbCluster/CrdbNode CRs, pods spawned across 3 zones. Full 3/3 Ready blocked by Docker Desktop CPU ceiling on this laptop, not a rename regression — belongs on cloud CI.

Two rounds of independent Opus review (no context sharing between them) — first round flagged 5 blockers + 10 non-blockers, all addressed. Second round flagged 4 blockers I missed on the first amend (all had the same shape: bare `./cockroachdb` in shell/markdown that the validate-rename.sh guard didn't cover). Guard extended to catch that class; second round's remaining items resolved.

## Follow-up work (post-merge cleanup)

**One-release-cycle fallbacks that MUST be deleted after this PR lands on master, tracked so they don't rot in the repo:**

1. **`.github/workflows/ci.yaml:328-345`** — the `chart-compatibility-e2e` job reads base chart versions from FETCH_HEAD via `git show`. Because the PR base commit still uses `cockroachdb-parent/…`, the workflow tries `cockroachdb-operator/…` first and falls back to `cockroachdb-parent/…` when the new path doesn't exist yet. Delete both fallback branches (`if [[ -z ... ]]; then base_*_version="$(git show FETCH_HEAD:cockroachdb-parent/…)"; fi`) once no open PR still has a pre-rename `master` as its base — typically ~1 release cycle after this PR merges. In-repo TODO marker at ci.yaml:328.

2. **`scripts/validate-rename.sh:38`** — the `':!.github/workflows/ci.yaml'` pathspec exclusion in guard #1 exists solely to let the fallback in item 1 above coexist with the guard. Drop this exclusion at the same time as item 1.

Everything else that references `cockroachdb-parent` in the tree (CHANGELOG historical headings, MIGRATION_v1alpha1_to_v1beta1.md historical checkout paths) is intentional documentation and stays forever.

## Test plan

- [ ] CI green
- [ ] Release Eng confirms no external automation (TeamCity / release bot) hardcodes `cockroachdb-parent/` for tag generation
- [ ] Cloud e2e (`tests/e2e/operator/...`) green
- [ ] Manual `make validate/rename` with k3d smoke on a machine with adequate CPU
- [ ] Follow-up PR opened to drop the fallbacks listed above after this lands on master


NishanthNalluri:
In general, I don't see any issues with this PR, since the changes do not modify the upstream chart names and are entirely focused on changing folder names in the repository. So this PR looks good to me.

But i just want to triple-check everything to leave no room for error, since even a minor oversight could break existing automation. Some things I would be cautious about with these changes are:

- Check that no Make commands, local scripts, or Git workflows are affected.
- Check that there is no impact on users who manage their Helm releases by cloning the git repo rather than pulling from upstream (I think this should be fine, since our compatibility E2E tests pull from upstream and perform upgrades using local charts)
- Ensure that any charts published after these changes do not introduce issues when users upgrade from previous versions while managing their releases entirely through upstream charts.

NishanthNalluri (approved):
LGTM

VishalJaishankar:
Rebased on latest master (picked up the operator v1.1.0 release commits). No conflicts, but one stale `cockroachdb-parent/` reference in a README section that got added in master — fixed and regenerated.

On the three things you flagged @NishanthNalluri:

1. Make targets, dev scripts, and workflow files all point at the renamed folders. `helm lint` passes on all three charts, `helm dependency update` resolves the subcharts cleanly, and `go test ./build/... ./tests/template/...` is green.

2. For folks who install by cloning the repo — the chart-compatibility-e2e test already does exactly that flow: it installs the previous published chart from upstream and then upgrades using the local working-tree charts. That local upgrade uses `operator.HelmChartPaths()`, which I updated to return the new paths.

3. For upstream-only upgrades — the two published subchart `name:` fields (`cockroachdb-chart`, `cockroachdb-operator-chart`) are untouched, the subchart aliases in the umbrella (`cockroachdb`, `operator`) are untouched, and rendering the umbrella still produces `helm.sh/chart: cockroachdb-<version>` on every resource. So an upstream user upgrading past this rename shouldn't see any pod churn. The one visible change is the `helm list` CHART column flipping from `cockroachdb-parent-*` to `cockroachdb-operator-*` if they installed the umbrella from a git clone — called out in the CHANGELOG.

Also ran `scripts/validate-rename.sh --skip-k3d` (static guards + generate idempotency + helm lint + go tests). All green.

